PR #1, opened 2021-12-01
initial burst metadata reader

LiangJYu:
### Features:
- Create ISCE3-compatible Sentinel1 burst class given S1 SAFE, subswath index, and polarization.
- Monotonically increasing bursts IDs.

### Install:

1. Set up and activate virtual environment with ISCE3.
2. Clone repository. 
```
$ cd ~/src
$ git clone https://github.com/LiangJYu/sentinel1-reader.git
```
3. Install into virtual environment with pip. From clone directory:
```
$ cd sentinel1-reader
$ pip install .
```

### Usage:
To process a single burst from a S1 SAFE zip and print burst ID + center lon/lat, from sentinel1-reader/bin directory:
```
$ python s1_read.py <path to S1 SAFE zip> <subswath index> <polarization>
```

### Todo:
- ~~Figure out why pip install is not working.~~


hfattahi (commented):
Thanks @LiangJYu. Very exciting that we are getting very close to have this reader work. I have not made it through the PR and test extensively. But here is few minor comments mainly suggestions on names of the class members. I will get to the rest of the PR soon.

sentinel1_reader/sentinel1_reader.py
```diff
@@ -0,0 +1,331 @@
...
+        # create masks for lines in current burst
+        mask0 = lines==ln0
+        mask1 = lines==ln1
+
+        # reverse order of 2nd set of points so plots of boundaries
+        # are not connected by a diagonal line
+        burst_lons = np.concatenate((lons[mask0], lons[mask1][::-1]))
+        burst_lats = np.concatenate((lats[mask0], lats[mask1][::-1]))
+
+        center_lon = np.mean(burst_lons)
```
vbrancat, line 176: Displaying my ignorance here :D. Are there cases in which the burst geometry might not be regular (e.g. rectangle) and in these cases would it make sense to replace the mean with the centroid to compute "the centre" of the burst?

sentinel1_reader/sentinel1_reader.py
```diff
@@ -0,0 +1,331 @@
...
+        Function used to open annotation file.
+
+    Returns:
+    --------
+    bursts : list
+        List of Sentinel1BurstSlc objects found in annotation XML.
+    '''
+    _, tail = os.path.split(annotation_path)
+    platform_id, subswath_id, _, pol = [x.upper() for x in tail.split('-')[:4]]
+    burst_interval = 2.758277
```
vbrancat, line 205: Can you write a comment on what we indicate as burst interval and how it is computed?

vbrancat (commented):
Thank you so much @LiangJYu . I left some minor comments and I hope to try the PR out asap. I was wondering if you have some Jupiter notebook displaying the usage of the reader. Might be worth uploading them in a folder in this repo.

sentinel1_reader/sentinel1_burst_slc.py
```diff
@@ -0,0 +1,190 @@
...
+    slant_range_time: float
+    starting_range: float
+    range_sampling_rate: float
+    range_pxl_spacing: float
+    shape: tuple()
+    azimuth_fm_rate: isce3.core.Poly1d
+    doppler: Doppler
+    range_processing_bandwidth: float
+    pol: str # {VV, VH}
+    id_str: str # t{track_number}_iw{1,2,3}_{burst_index}
```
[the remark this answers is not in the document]
LiangJYu: `id` is a built in function. Would the more verbose `identification` be ok?

sentinel1_reader/sentinel1_burst_slc.py
```diff
@@ -0,0 +1,190 @@
...
+    platform_id: str # S1{A,B}
+    center: tuple # {center lon, center lat} in degrees
+    border: list
+    # VRT params
+    # TODO maybe make these own dataclass?
```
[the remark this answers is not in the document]
LiangJYu: Haven't thought much about after writing the comment. You have an opinion?

sentinel1_reader/sentinel1_burst_slc.py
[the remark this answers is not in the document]
hfattahi: How about "burst_id" or "unique_id". 

@vbrancat @yunjunz do you have suggestions here. Note that this is a member of the "burst" class so the usage is usually like 

```
burst.busrt_id
``` 
or 
```
burst.unique_id
```

sentinel1_reader/sentinel1_burst_slc.py
```diff
@@ -0,0 +1,190 @@
...
+
+        # determine start and end time from metadata
+        pulse_length = (self.shape[0] - 1) * self.azimuth_time_interval
+        t_pulse_end = self.sensing_start + datetime.timedelta(seconds=pulse_length)
+
+        # find files with self.platform_id
+        item_valid = lambda item, sat_id: os.path.isfile(item) and sat_id in item
+        orbit_files = [item for item in os.listdir(orbit_dir)
+                       if item_valid(f'{orbit_dir}/{item}', self.platform_id)]
+        if not orbit_files:
```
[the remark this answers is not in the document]
hfattahi: Good point @vbrancat I was actually thinking the same. But then I was hesitant if we want to have the reader to be isolated as a pure reader and let workflows at higher level be worried about downloading orbits and all the junk around (connection failure etc). I don't have a strong opinion really but I just wanted to make sure you also note that point. 
Anyways I think the reader should accept external orbit and that should be the default behavior.  
A middle ground would be to have a flag for orbit_download which is off by default. So if the orbit is not found and if the user specifically asks to download, then the reader attempts to download.

sentinel1_reader/sentinel1_burst_slc.py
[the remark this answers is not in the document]
yunjunz: `burst_id` sounds good to me.

sentinel1_reader/sentinel1_reader.py
```diff
@@ -0,0 +1,331 @@
...
+    Returns:
+    ------
+    _ : datetime.datetime
+        datetime.dateime object parsed from given time string.
+    '''
+    return datetime.datetime.strptime(t_str, fmt)
+
+def parse_polynomial_element(elem, poly_name):
+    '''
+    Parse az fm rate element to reference time and poly1d.
```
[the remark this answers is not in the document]
LiangJYu: `az` to `azimuth`?

What about `fm`? `FM` or `frequency modulation`?

sentinel1_reader/sentinel1_burst_slc.py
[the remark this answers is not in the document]
hfattahi: I think its fine as is

sentinel1_reader/sentinel1_reader.py
LiangJYu: I assume you're referring to mean vs centroid like [this](https://github.com/Turfjs/turf/issues/334)?

sentinel1_reader/sentinel1_reader.py
```diff
@@ -0,0 +1,331 @@
...
+    _ : LUT2d
+        LUT2d calculated from poly1d.
+    '''
+    (n_lines, n_samples) = shape
+    # calculate all slant ranges in grid
+    slant_ranges = starting_slant_range + np.arange(n_samples) * slant_range_res
+
+    # no az dependency, but LUT2d required, so ensure all az coords covered
+    # TODO fix magic number of 2 days minus sensing_start
+    az_times = 2*24*3600 + np.array([0, n_lines * az_time_interval])
```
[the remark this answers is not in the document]
LiangJYu: @hfattahi previously suggested 2 days before sensing start as a reference epoch in CSLC workflow demo development. It worked and I simply stuck with it.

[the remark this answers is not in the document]
LiangJYu: @vbrancat Do you think the following is sufficient documentation?
```suggestion
     # offset by -2 days in seconds (referenece epoch)
     offset_ref_epoch = 2 * 24 * 3600
    az_times = offset_ref_epoch + np.array([0, n_lines * az_time_interval])
```

sentinel1_reader/sentinel1_burst_slc.py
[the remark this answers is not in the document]
vbrancat: `burst_id` sounds good to me. Somewhere (maybe in a comment or in the description of the package) we should specify and clarify the "uniqueness"

sentinel1_reader/sentinel1_burst_slc.py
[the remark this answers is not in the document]
vbrancat: I agree. The higher-level workflow should **properly** handle this (e.g. check the internet connection and all this stuff). However, I do not see any harm in providing an optional functionality to download the orbit (if not present). Maybe the functionality (it can be a simple flag) should not have too much logic in it (e.g. not checking the internet connection). I will let @LiangJYu illuminate us if this solution does not violate the "separation of functionalities" principle (https://en.wikipedia.org/wiki/Single-responsibility_principle)

sentinel1_reader/sentinel1_reader.py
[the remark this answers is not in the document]
vbrancat: FM (Frequency Modulation) :P

sentinel1_reader/sentinel1_reader.py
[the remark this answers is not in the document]
vbrancat: Looks good

sentinel1_reader/sentinel1_reader.py
vbrancat: Yes! I guess the two concepts should degenerate to the same thing for regular geometries (e.g. rectangles). This is not necessarily the case for irregular geometries. However, I think burst geometries should be consistently regular, so I guess having mean or centroid should not matter. Am I wrong?

LiangJYu: I don't think we'll encounter any irregular geometries so I agree with you.

What this doesn't address is rectangle intersects the date line. Do you know of a way to handle this?

vbrancat: Excellent point. If we generate the burst polygon in lat/lon, we will run into the crossing date line issue. I think the correct way to handle it would be:
1. Check if the burst polygon crosses the dateline
2. If it does, split the polygon in two parts, one polygon before covering the burst area before the dateline, and another polygon covering the area after the dateline. 

What do you think?

setup.cfg
```diff
@@ -0,0 +1,18 @@
...
+    Development Status :: 2 - Pre-Alpha
+    Intended Audience :: Science/Research
+    Programming Language :: Python :: 3
+    License = file : LICENSE
+    Operating System :: OS Independent
+
+[options]
+package_dir =
+packages = find:
+python_requires = >=3.6
```
yunjunz, line 18:
```suggestion
python_requires = >=3.7
```

Because the `dataclasses` module is introduced in Python 3.7 and used.

sentinel1_reader/sentinel1_reader.py
yunjunz: The link below might be useful to add as well:
```
For IW mode, one burst has a duration of ~2.75 seconds and a burst overlap of approximately ~0.4 seconds.
https://sentinels.copernicus.eu/web/sentinel/user-guides/sentinel-1-sar/product-types-processing-levels/level-1
```

sentinel1_reader/sentinel1_reader.py
LiangJYu: I will use [`check_dateline`](https://github.com/isce-framework/isce3/blob/e2c76c70c304a52fe583a0dd4e0455eab12475d9/python/packages/nisar/workflows/stage_dem.py#L39) from stage_dem.py as a guide.

sentinel1_reader/sentinel1_burst_slc.py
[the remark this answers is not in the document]
hfattahi: ok with "burst_id" as well.

sentinel1_reader/sentinel1_burst_slc.py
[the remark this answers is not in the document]
hfattahi: Sounds good to me as far as it is optional to use.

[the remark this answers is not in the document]
LiangJYu: Especially when running in production, I also agree the higher-level workflow should download the orbits. For other use cases, I'm not opposed to including a download option. Perhaps something like the following?
```python
def get_local_orbit_file(t_start, t_stop, orbit_dir):
    # try to find and return path file in orbit_dir else return empty string

def get_online_orbit_file(t_start, t_stop):
    # try to find and return file from online source else return empty string

def get_isce3_orbit(..., online_ok=False):
    ...
    # try local first
    orbit_file = get_local_orbit_file(...)

    # try online if nothing local and ok to go online
    if not orbit_file and online_ok:
        orbit_file = get_online_orbit_file(...)

        # check if nothing found online
        if not orbit_file:
            raise ValueError("no orbit file found")
    ...
```

sentinel1_reader/sentinel1_burst_slc.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Returns:
+        --------
+        _ : datetime
+            Sensing mid as datetime object.
+        '''
+        d_seconds = 0.5 * (self.shape[0] - 1) * self.azimuth_time_interval
+        return self.sensing_start + datetime.timedelta(seconds=d_seconds)
+
+    def get_isce3_orbit(self, orbit_dir: str):
```
LiangJYu, line 126: I think it makes sense to load orbit while the XML is read and to make it a member of `Sentinel1BurstSlc`. It'll be highly unlikely that a swath will span multiple orbit files, so repeated calls current function is doing a lot repeated/redundant file reading.

sentinel1_reader/sentinel1_burst_slc.py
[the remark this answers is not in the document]
hfattahi: This looks good to me.

sentinel1_reader/sentinel1_burst_slc.py
hfattahi: I agree that orbit may be a member of sentinel-1 burst SLC. But still you need a method to return the orbit as we will need orbit for workflows as you know.
The extra call to get_orbit did not bother me. actually it's more clear what is going on.

bin/s1_read.py
```diff
@@ -9,8 +9,8 @@
     if i_subswath < 1  or i_subswath > 3:
         raise ValueError("i_subswath not <1 or >3")
     pol = sys.argv[3]
-    if pol not in ['vv', 'vh']:
-        raise ValueError("polarization not 'vv' or 'vh'")
+    if pol not in ['vv', 'vh', 'hh']:
```
hfattahi, line 12: please add 'hv' as well

bin/s1_read.py
```diff
@@ -9,8 +9,8 @@
...
+        raise ValueError("polarization not in {'vv', 'vh', 'hh'}")
```
hfattahi, line 13: add 'hv' as well please

sentinel1_reader/sentinel1_burst_slc.py
```diff
@@ -18,25 +18,24 @@ class Sentinel1BurstSlc:
...
-    range_pxl_spacing: float
+    range_pixel_spacing: float
     shape: tuple()
     azimuth_fm_rate: isce3.core.Poly1d
     doppler: Doppler
-    range_processing_bandwidth: float
-    pol: str # {VV, VH}
-    id_str: str # t{track_number}_iw{1,2,3}_{burst_index}
+    range_bandwidth: float
+    polarization: str # {VV, VH, HH}
```
hfattahi, line 33: add  HV to the comment

LiangJYu:
After the most recent commits, I think the interfaces will be changing much so this will be ready to use in workflow development.

Orbit download from ASF not implemented. Is it immediately required for workflow development?

vbrancat:
@LiangJYu, orbit downloading in the sentinel reader are not necessary at the moment. They are "nice to have" but they can be added later. Plus, we will have the S1 CSLC managing that. Let's get the `s1_reader` merged and start working on a  draft of the S1 CSLC workflow.

sentinel1_reader/sentinel1_reader.py
```diff
@@ -0,0 +1,387 @@
...
+        mask1 = lines==ln1
+
+        # reverse order of 2nd set of points so plots of boundaries
+        # are not connected by a diagonal line
+        burst_lons = np.concatenate((lons[mask0], lons[mask1][::-1]))
+        burst_lats = np.concatenate((lats[mask0], lats[mask1][::-1]))
+
+        poly = shapely.geometry.Polygon(zip(burst_lons, burst_lats))
+        center_pts[i] = poly.centroid.xy
+        boundary_pts[i] = check_dateline(poly)
```
yunjunz, line 219: Shouldn't we apply the `dateline` checking before calculating the centroid? One simple way could be just to use the largest polygons among the returned list of polygons.

LiangJYu:
@yunjunz Do have a preference on the repository directory/file structure? Do you prefer something closer to you've done in [this PR](https://github.com/opera-adt/sentinel1-burst-id/pull/4)?

sentinel1_reader/sentinel1_reader.py
yunjunz: Maybe @vbrancat or @hfattahi knows better, I would think we want only one `center_pts` for each burst, so we only need the largest polygon, instead of all the spited ones.

yunjunz:
@LiangJYu I do like the `src/{module_name}` structure more, it would be great if we could keep the same style here too.

sentinel1_reader/sentinel1_reader.py
LiangJYu: Why would the center of the largest split poly be preferred to center of the burst? IIRC polygon splitting was to accommodate DEM downloading?

BTW both `Polygon.centroid` and the previous averaging of points do a poor job of calculating the actual center of the burst when its split by the dateline. AFAICT using `Polygon.centroid` as-is gives wrong results. The averaging method is a bit off. The following seems to work better:

1. convert LLH to XYZ
2. average XYZ
3. covert average XYZ to LLH
4. set average LLH height to 0

hfattahi: Interesting. I think it should work!

yunjunz: I like the XYZ calculation too.

hfattahi:
@LiangJYu would you please update the README with the instruction that you have in the PR description above. 
Also you may add few lines of example as you had in a script. e.g.:

```
from sentinel1_reader import sentinel1_reader, sentinel1_orbit_reader

zip_path = "S1A_IW_SLC__1SDV_20190909T134419_20190909T134446_028945_03483B_B9E1.zip"
i_subswath = 2
pol = "HH"

#returns the list of the bursts
bursts = sentinel1_reader.zip2bursts(zip_path, osv_list, i_subswath, pol)

```

src/sentinel1_reader/sentinel1_reader.py
```diff
@@ -0,0 +1,418 @@
...
+               n_subswath: int, pol: str):
+    '''
+    Find bursts in a Sentinel 1 zip file
+
+    Parameters:
+    -----------
+    zip_path : str
+        Path the zip file.
+    osv_list : xml.etree.ElementTree.Element
+        ElementTree containing orbit state vectors
```
hfattahi, line 397: @LiangJYu I find this argument a bit strange. zip2burst is the interface with the user of the reader and it looks very user unfriendly to me to require an argument which is xml.etree.ElementTree.Element!
Is there any reason you ended up with this approach?  can we avoid this. Shall we simply require an orbit file and not an orbit directory so that the user of the reader will be responsible to pass the correct orbit file.

hfattahi: Actually I think the reader should just simply take the orbit file and not the orbit directory. As discussed [here](https://github.com/opera-adt/COMPASS/discussions/3) it should be the responsibility of the user of the S1 reader at higher level to find the right orbit file and pass it to the workflows such that the reader can use.

LiangJYu: I agree with you it's currently not as user friendly. I opted to do this so the same orbit XML doesn't need to be reread in the event we want to process all bursts/all polarizations for a zip. This is slightly faster (<< 1 sec to read the XML) but not significantly so in grand scheme of things.

I like the idea of the reader take in the correct orbit file. We have or will have tools available to help users pick the correct file. The orbit reader can pick the correct file (if it exists) from a directory. The same logic can be fairly easily extended to scrape the ASF orbit site.

hfattahi: If the improvement is << 1 sec, then I suggest to not sacrifice the user friendliness. 
Yes, I think the reader should expect the correct orbit file and not orbit directory.

.gitignore
```diff
@@ -0,0 +1,131 @@
+*.DS_Store
```
vbrancat, line 1: Maybe we do not need the `.gitignore`

src/sentinel1_reader/sentinel1_reader.py
```diff
@@ -0,0 +1,418 @@
...
+                                      range_sampling_rate, range_pxl_spacing,
+                                      (n_lines, n_samples), az_fm_rate, doppler,
+                                      rng_processing_bandwidth, pol, burst_id,
+                                      platform_id, center_pts[i], boundary_pts[i],
+                                      orbit, tiff_path, i, first_valid_sample,
+                                      last_sample, first_valid_line, last_line)
+
+    return bursts
+
+def zip2bursts(zip_path: str, osv_list: ET.Element,
```
hfattahi, line 387: Sorry for nitpicking. Could we have a better name? How about "read_bursts" ? @vbrancat @yunjunz do you have other suggestions?

.gitignore
LiangJYu: I like having an ignore file as an easy way of ensuring ancillary/extraneous files that get created in the repo directory don't get committed into the project. e.g. Calling `git commit -am "some message"` when there's random non-source files in the repo.

Are there any specific reason(s) why your find the ignore file unnecessary?

src/sentinel1_reader/sentinel1_reader.py
```diff
@@ -0,0 +1,418 @@
...
+
+        first_valid_line = [x > 0 for x in first_valid_samples].index(True)
+        n_valid_lines = [x > 0 for x in first_valid_samples].count(True)
+        last_line = first_valid_line + n_valid_lines - 1
+
+        first_valid_sample = max(first_valid_samples[first_valid_line],
+                                 first_valid_samples[last_line])
+        last_sample = min(last_valid_samples[first_valid_line],
+                          last_valid_samples[last_line])
+        n_valid_samples = last_sample - first_valid_sample
```
gshiroma, line 374: Same comment as above. If `last_sample` is `3` and `first_valid_sample` is `1`, shouldn't `n_valid_samples` equal `3` rather than `2`?

LiangJYu:
All issues/comments have been addressed. If there's nothing else, I believe this PR is ready to be merged.

vbrancat:
@LiangJYu Would you mind fix the "codacy" issues before merging?

LiangJYu:
I'm giving up on addressing Codacy formatting issues. In acaa997, I addressed issues flagged from 9aed622 with suggestions from the 9aed622 Codacy summary. Codacy issues in acaa997 seem to conflict with issues and suggestions from 9aed622.

If we're going to do formatting checks, I much prefer using tools that can also be run on both GitHub and on the local development environment. This arrangement would reduce the number of commits attempting to appease the black box formatting test.

LiangJYu:
@yunjunz Thanks for addressing the formatting in the README! I'm surprised fe4df29 didn't encounter all the flags in Python that acaa997 encountered (see below).
image

yunjunz:
`Codacy` has a learning curve as well. For the flag, for example, I just disabled this pattern, so it won't bother us next time.

src/sentinel1_reader/sentinel1_reader.py
```diff
@@ -0,0 +1,412 @@
...
+    n_subswath : int
+        Integer of subswath of desired burst. {1, 2, 3}
+    pol : str
+        Polarization of desired burst. {HH, VV, HV, VH}
+
+    Returns:
+    --------
+    bursts : list
+        List of Sentinel1BurstSlc objects found in annotation XML.
+    '''
```
hfattahi, line 401: I suggest to add sanity check for n_subswath and pol right here? we know they can have very limited values. 
without those checks, the reader fails with non-intuitive messages such as this one if you ask for sub-swath 4:

```
bursts = sentinel1_reader.burst_from_zip(zip_path, orbit_path, 4, "vv")
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-20-c988331e584e> in <module>
----> 1 bursts = sentinel1_reader.burst_from_zip(zip_path, orbit_path, 4, "vv")

~/anaconda3/envs/isce3/lib/python3.9/site-packages/sentinel1_reader/sentinel1_reader.py in burst_from_zip(zip_path, orbit_path, n_subswath, pol)
    403     with zipfile.ZipFile(zip_path, 'r') as z_file:
    404         # find annotation file
--> 405         f_annotation = [f for f in z_file.namelist() if 'calibration' not in f and id_str in f and 'annotation' in f][0]
    406
    407         # find tiff file

IndexError: list index out of range
```

hfattahi (approved):
Thank you @LiangJYu. This looks great. The metadata seems correct and the extracted raster is identical to that extracted by isce2.  I have only few minor comments which should be easy to address.

vbrancat (approved):
LGTM 👍 . Thank you so much for have been so patience and having addressed all the issues

src/sentinel1_reader/sentinel1_burst_slc.py
```diff
@@ -0,0 +1,125 @@
...
+
+        Parameters:
+        -----------
+        out_path : string
+            Path of output VRT file.
+        '''
+        line_offset = self.i_burst * self.shape[0]
+
+        inwidth = self.last_valid_sample - self.first_valid_sample
+        inlength = self.last_valid_line - self.first_valid_line
```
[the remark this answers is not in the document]
gshiroma: I don't see this comment addressed yet. Please let me know if it makes sense or not. Thank you.

src/sentinel1_reader/sentinel1_orbit_reader.py
```diff
@@ -0,0 +1,67 @@
...
+
+    if not os.path.isdir(orbit_dir):
+        raise NotADirectoryError(f"{orbit_dir} not found")
+
+    # date format used in file names
+    fmt = "%Y%m%dT%H%M%S"
+
+    # determine start and end times from file name
+    file_name_tokens = os.path.basename(zip_path).split('_')
+    platform_id = file_name_tokens[0]
```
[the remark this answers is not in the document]
gshiroma: Same here. Please, let me know if you think the check is not needed.

gshiroma (commented):
Hi @LiangJYu , I think my comments were not addressed yet. Maybe there is some commit left to push?

src/sentinel1_reader/sentinel1_orbit_reader.py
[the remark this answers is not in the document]
LiangJYu: See lines 35-37 below.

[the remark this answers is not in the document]
LiangJYu: Add in 7de0f9a

LiangJYu:
@gshiroma width and length addressed [here](https://github.com/opera-adt/sentinel1-reader/blob/277c29b56f9fdb18c25804b07ab049e5ee7308ca/src/sentinel1_reader/sentinel1_burst_slc.py#L83-L84) with 20bc7ad 

Were you expecting something else?

gshiroma (approved):
I see the changes now. I don't know why GitHub was not showing me your updates previously. Thank you for addressing all my comments. Nice update, @LiangJYu ! LGTM!

PR #8, opened 2022-01-15
Adding azimuth carrier computation for bursts

vbrancat:
This PR adds the functionality of computing the azimuth carrier polynomials to the S1 reader for each burst (if requested).

This functionality will be needed for both COMPASS range/Doppler and geocoded CSLC workflow. To avoid code duplication, it makes sense that this functionality sits in the reader.

Note: the `polyfit` function in `sentinel1_burst_slc.py `needs help for the documentation.

### DIFF
--- a/src/sentinel1_reader/sentinel1_burst_slc.py
+++ b/src/sentinel1_reader/sentinel1_burst_slc.py
@@ -1,4 +1,4 @@
-import datetime as datetime
+import datetime
 from dataclasses import dataclass
 
 import isce3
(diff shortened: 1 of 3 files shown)


hfattahi (approved):
Thanks @vbrancat for pushing this part of the reader.

LiangJYu:
LGTM. Just address Codacy warnings.

yunjunz:
The `datetime` module is imported twice in the code with different styles, would recommend unifying them, such as below at the top of the script:

```python
import datetime as dt
```

@LiangJYu We has similar syle inconsistency of `datetime` in `sentinel1_orbit_reader.py` and `sentinel1_reader.py` as well, might be a good idea to use the same import style as above as well?

vbrancat:
@LiangJYu did my best to address codacy issues. There is only one remaining related to an unused (but in reality is used) variable. @yunjunz you are right. Our imports of the `datetime` module are ambiguous. However, we cannot use your suggestion (`import datetime as dt`) as we use `dt` in to indicate a time difference. I decided to resolve the issue you pointed out using `import datetime as datetime`. It is a bit long but rather straightforward and unambiguous and it also help code readability. Waiting for your green light to merge the PR.

yunjunz (approved):
Looks great to me!

PR #11, opened 2022-01-27
added output burst as GTiff

LiangJYu:
This PR adds capability of writing a burst SLC as GTiff.

### DIFF
--- a/bin/s1_read.py
+++ b/bin/s1_read.py
@@ -27,3 +27,6 @@ if __name__ == "__main__":
 
     for i, burst in enumerate(bursts):
         print(burst.burst_id, burst.center)
+    burst.slc_to_file('burst.slc')
+    burst.slc_to_file('burst.tif', 'GTiff')
+    burst.slc_to_file('burst.vrt', 'VRT')
--- a/bin/s1_read.py
+++ b/bin/s1_read.py
@@ -4,6 +4,7 @@ import sys
 from sentinel1_reader import sentinel1_reader, sentinel1_orbit_reader
 
 if __name__ == "__main__":
+    '''testing script that prints burst info and write SLC to file'''
     # TODO replace with argparse
     zip_path = sys.argv[1]
     if not os.path.isfile(zip_path):
@@ -25,8 +26,13 @@ if __name__ == "__main__":
 
     bursts = sentinel1_reader.burst_from_zip(zip_path, orbit_path, i_subswath, pol)
 
+    # print out IDs and lat/lon centers of all bursts
     for i, burst in enumerate(bursts):
         print(burst.burst_id, burst.center)
+
+    # write to ENVI (default)
     burst.slc_to_file('burst.slc')
+    # write to geotiff
     burst.slc_to_file('burst.tif', 'GTiff')
+    # write to VRT
     burst.slc_to_file('burst.vrt', 'VRT')
--- a/src/sentinel1_reader/sentinel1_burst_slc.py
+++ b/src/sentinel1_reader/sentinel1_burst_slc.py
@@ -220,7 +220,7 @@ class Sentinel1BurstSlc:
                                                  width,
                                                  ref_epoch)
 
-    def slc_to_gtiff(self, out_path):
+    def slc_to_file(self, out_path: str, fmt: str = 'ENVI'):
         '''Write burst to GTiff file.
 
         Parameters:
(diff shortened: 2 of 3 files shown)


LiangJYu:
I don't think Codacy's issues can be resolved. The unused variable `dst_ds` is being used. The assignment to `None` flushes to disk.

Below is the offending snippet:
```python
        # open temporary VRT and translate to GTiff
        src_ds = gdal.Open(temp_vrt.name)
        dst_ds = gdal.Translate(out_path, src_ds, format='GTiff')

        # clean up
        src_ds = None
        dst_ds = None
```

src/sentinel1_reader/sentinel1_burst_slc.py
```diff
@@ -218,7 +220,31 @@ def as_isce3_radargrid(self):
                                                  width,
                                                  ref_epoch)
 
-    def to_vrt_file(self, out_path):
+    def slc_to_gtiff(self, out_path):
```
hfattahi, line 223: this can be general. Does not need to be GeoTiff. Can be any gdal format with default to ENVI.
I suggest to change the name of the function to 
write_to_disk or write_sls or write or even a better name if you or others can think of one? 
then have the output format as an input argument default to ENVI.

src/sentinel1_reader/sentinel1_burst_slc.py
```diff
@@ -218,7 +220,31 @@ def as_isce3_radargrid(self):
...
+        temp_vrt = tempfile.NamedTemporaryFile(dir=dst_dir, suffix='.tif')
+        self.slc_to_vrt_file(temp_vrt.name)
+
+        # open temporary VRT and translate to GTiff
+        src_ds = gdal.Open(temp_vrt.name)
+        dst_ds = gdal.Translate(out_path, src_ds, format='GTiff')
+
+        # clean up
+        src_ds = None
+        dst_ds = None
```
yunjunz, line 244: I don't use gdal very frequently, just curious that regarding the codacy flag, can we change these lines to below?

```python
gdal.Translate(out_path, src_ds, format='GTiff')

# clean up
src_ds = None
```

bin/s1_read.py
```diff
@@ -27,3 +27,6 @@
 
     for i, burst in enumerate(bursts):
         print(burst.burst_id, burst.center)
+    burst.slc_to_file('burst.slc')
+    burst.slc_to_file('burst.tif', 'GTiff')
```
hfattahi, line 31: GTiff may not be our favorite format here. First the bursts are not geocoded so using GeoTiff may not be needed. Second we commonly use mdx to display the SLC data and mdx is happy with flat binary files. Using ENVI will make a simple flat binary file and that's why I prefer until we have a more advanced viewer working as fast as mdx.

LiangJYu: I forgot to write comments. Fixed in 06b636e.

I included the 3 calls to `slc_to_file` at the bottom on `s1_read.py` to test if the 3 write modes worked. https://github.com/opera-adt/sentinel1-reader/blob/06b636e50e21170310f52444f4264bc697c108fd/bin/s1_read.py#L33-L38.

yunjunz (approved):
Looks great to me.

hfattahi:
I just tried the reader in this PR to make sure it works. It worked on S1A but failed on a new S1B SLC. 
It looks like in case of S1B the annotation_path is populated incorrectly:
S1B_IW_SLC__1SDV_20211115T134338_20211115T134405_029599_03884F_2E66.SAFE/annotation/```rfi/rfi```-s1b-iw2-slc-vv-20211115t134339-20211115t134404-029599-03884f-005.xml

The problem is that the newer Sentinel-1 SLCs may have ```rfi``` subfolder which also includes xml files. We need to exclude ```rfi``` from [here](https://github.com/LiangJYu/sentinel1-reader/blob/slc_to_envi/src/sentinel1_reader/sentinel1_reader.py#L415) by replacing the line with the following:

```
f_annotation = [f for f in z_file.namelist() if 'calibration' not in f and 'rfi' not in f and id_str in f and 'annotation' in f]
```
I think this solves the issue for now until they add another subfolder. It would be good to do this search more rigorously so that adding additional subfolders in future won't break our annotation file searing. I like the approach we use [here](https://github.com/isce-framework/isce2/blob/main/components/isceobj/Sensor/TOPS/Sentinel1.py#L222). How about the same approach to be adopted in this reader as well. In other words we would use the following:

```
swathid = 's1?-iw%d'%(i_subswath)
with zipfile.ZipFile(zip_path, 'r') as z_file:
       # find annotation file
       pattern = os.path.join('*SAFE','annotation', swathid) + '-slc-' + pol + '*.xml'
       f_annotation = fnmatch.filter(z_file.namelist(), pattern)

``` 
 This way the reader will always find the annotation file even if more subfolders added later and we don't need to exclude rfi, calibration etc.

hfattahi (approved):
LGTM!

PR #15, opened 2022-02-04
README: switch to `pol=VV`

hfattahi:
In the example we have provided a Dual-pol Vertical (DV) zip file but we passed ```pol=HH``` which does not exist in the zip file. Updating to ```pol=VV``` so that the example makes sense.


yunjunz (approved):
LGTM.

LiangJYu (approved):
LGTM

vbrancat (approved):
LGTM

Issue #10, opened 2022-01-20
unit test for the sentinel1-reader


hfattahi:
This repository still lacks a unit test. I think writing the test is easy if we could have a Sentinel-1 SLC. One Sentinel-1 SLC is too big (4+ GB ) to store in the repo. We could download the data as part of the CI script but the problem is ASF and Scihub require credentials to download the data. What is the right approach to tackle this issue? Is there any sample Sentinel-1 SLC on ESA website that we could pull without credentials?

If there is no way around the data download issue, an alternative would be to upload only the xml files of one SLC to the repo, emulate the SAFE file data structure, zip it and use it for the test. 

@yunjunz @LiangJYu @vbrancat @gshiroma @rtburns-jpl  can you think of any other solution?

yunjunz:
In circle CI, we could put the credentials as "Environment Variables", and pass it to the script, the value of the variables cannot be read or edited in the app once they are set.

hfattahi:
> In circle CI, we could put the credentials as "Environment Variables", and pass it to the script, the value of the variables cannot be read or edited in the app once they are set.

Awesome Yunjun. So that should solve the issue I guess. unless if we don't want to have the CI spend a lot of time downloading the entire SLC while we only need the metadata. 

Perhaps we could have two tests, one that we would download the SLC and that could be a nightly/weekly test and a faster test which we could download only the xml files (e.g. using [this](https://github.com/avalentino/asfsmd)) and as I suggested create the SAFE file data structure. This should be a much faster test to run.

vbrancat:
Sorry for chiming in too late. Most of the time I download S1-A/B data using AWS buckets (e.g. http://sentinel1-slc-seasia-pds.s3-website-ap-southeast-1.amazonaws.com/datasets/slc/v1.1/). This should completely eliminate the need for username and password credentials but it requires coding a bit on s3 bucket interaction.

hfattahi:
@vbrancat that's nice. I was not aware of this. So we can have multiple options for downloading data for a unit test!

hfattahi:
Thanks @vbrancat I was not aware of this s3 bucket.

vbrancat:
> > In circle CI, we could put the credentials as "Environment Variables", and pass it to the script, the value of the variables cannot be read or edited in the app once they are set.
> 
> Awesome Yunjun. So that should solve the issue I guess. unless if we don't want to have the CI spend a lot of time downloading the entire SLC while we only need the metadata.
> 
[...]

Issue #3, opened 2021-12-10
Abiltiy to download orbits from ASF


LiangJYu:
Currently all orbit files need to be local for reader to work. Eventually something using BeautifulSoup should suffice to parse [ASF](https://s1qc.asf.alaska.edu/aux_poeorb/).

LiangJYu:
@hfattahi as mentioned you [here](https://github.com/opera-adt/COMPASS/pull/29#issuecomment-1142578121), this [module](https://github.com/scottstanie/sentineleof) looks like it'll do exactly what we need. 

Besides adding the module to the [environment yaml](https://github.com/opera-adt/s1-reader/blob/main/environment.yaml), should the ability to download if one is unavailable be included?

Does it matter that the orbits are downloaded from scihub.copernicus.eu instead of ASF?

PR #18, opened 2022-02-07
Unit test to download S1-A/B SAFE file

vbrancat:
This PR adds a unit test to download S1-A/B data. To avoid using credential info, data are downloaded from the AWS s3 seasia bucket containing S1-A/B data covering only South East Asia.

The PR should partially address issue #10. Another PR will follow for a test downloading the XML file and test  various reader functionalities.

Note: To download approx 4 GB of data takes around 10-15 min on average. It is recommended that this test runs nightly as a part of the CI suite.

### DIFF
--- a/tests/download_data.py
+++ b/tests/download_data.py
@@ -1,0 +1,31 @@
+import os
+import shutil
+from urllib.request import urlopen
+
+s3_url = "http://sentinel1-slc-seasia-pds.s3-website-ap-southeast-1.amazonaws.com/datasets/slc/v1.1"
+data_info = "2021/08/15"
+granule_id = "S1A_IW_SLC__1SDV_20210815T100025_20210815T100055_039238_04A1E8_D145"
+
+
+def download_granule(in_url):
+    zip_file = in_url.split('/')[-1]
+    if os.path.isfile(zip_file) == False:
+        print(f'Start downloading of {in_url}')
+        with urlopen(in_url) as response, open(zip_file, 'wb') as ofile:
+            shutil.copyfileobj(response, ofile)
+
+
+def test_data_download():
+    '''
+    Test S1-A/B data download. Use seasia AWS s3 bucket
+    to download data. Note, it contains only data acquired
+    South East Asia, Taiwan, Korea, and Japan.
+    '''
+    # Get Full data URL
+    data_url = f'{s3_url}/{data_info}/{granule_id}/{granule_id}.zip'
+    download_granule(data_url)
+    print('Done')
+
+
+if __name__ == "__main__":
+    test_data_download()
--- a/tests/download_data.py
+++ b/tests/download_data.py
@@ -1,6 +1,4 @@
 import os
-import shutil
-from urllib.request import urlopen
 
 s3_url = "http://sentinel1-slc-seasia-pds.s3-website-ap-southeast-1.amazonaws.com/datasets/slc/v1.1"
 data_info = "2021/08/15"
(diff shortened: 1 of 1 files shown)


tests/download_data.py
```diff
@@ -0,0 +1,31 @@
...
+import shutil
+from urllib.request import urlopen
+
+s3_url = "http://sentinel1-slc-seasia-pds.s3-website-ap-southeast-1.amazonaws.com/datasets/slc/v1.1"
+data_info = "2021/08/15"
+granule_id = "S1A_IW_SLC__1SDV_20210815T100025_20210815T100055_039238_04A1E8_D145"
+
+
+def download_granule(in_url):
+    zip_file = in_url.split('/')[-1]
```
rtburns-jpl, line 11:
```suggestion
    zip_file = os.path.basename(in_url)
```

tests/download_data.py
```diff
@@ -0,0 +1,31 @@
...
+    if os.path.isfile(zip_file) == False:
```
rtburns-jpl, line 12:
```suggestion
    if not os.path.isfile(zip_file):
```

rtburns-jpl (approved):
Nice, started a download locally and it seems to be working. The download is pretty big and doesn't show any sign that it is making progress, do you know if there's an easy way to show a progress bar?

yunjunz:
`wget` will print out a nice progress bar as below, although we will have to use the `os.system(wget)` syntax for this.

```
wget https://zenodo.org/record/5498198/files/FernandinaSenDT128.tar.xz
--2022-02-06 04:09:35--  https://zenodo.org/record/5498198/files/FernandinaSenDT128.tar.xz
Resolving zenodo.org (zenodo.org)... 137.138.76.77
Connecting to zenodo.org (zenodo.org)|137.138.76.77|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 277140032 (264M) [application/octet-stream]
Saving to: ‘FernandinaSenDT128.tar.xz’

FernandinaSenDT128. 100%[===================>] 264.30M  10.4MB/s    in 34s     

2022-02-06 04:10:10 (7.79 MB/s) - ‘FernandinaSenDT128.tar.xz’ saved [277140032/277140032]
```

LiangJYu:
SO suggestions:
[`shutil` wrapper](https://stackoverflow.com/a/29967714)
[add/use `requests`](https://stackoverflow.com/a/15645088)

I'm partial to the `shutil` wrapper to keep dependencies light.

rtburns-jpl:
Some distros don't come with wget, but it's a pretty small dependency to add. Sounds like a good approach to me 👍

vbrancat:
@rtburns-jpl before adding a `wget` dependency I can try to see if there is a workaround with `requests`

vbrancat:
Another option (lighter in terms of code to implement) is to have as a dependency `python-wget`

Using:
wget.downloat(in_url)

gives this nice progress bar
Screen Shot 2022-02-07 at 2 33 24 PM

@yunjunz @rtburns-jpl @LiangJYu how does that sound?

yunjunz:
I am not sure, the `python-wget` has not been updated since 2015 and it seems to be abanded by the developer, thus, I am more inclined to use `wget` directly. 

Note that we should use a different dependency list for testing, which is probably longer than the code dependency that end-users need, so I am not very concerned about adding `wget`.

vbrancat:
> I am not sure, the `python-wget` has not been updated since 2015 and it seems to be abanded by the developer, thus, I am more inclined to use `wget` directly.
> 
> Note that we should use a different dependency list for testing, which is probably longer than the code dependency that end-users need, so I am not very concerned about adding `wget`.

Sounds good :)

LiangJYu:
@yunjunz's solution `wget` solution sounds good to me.

Since you've already written this [function](https://github.com/vbrancat/sentinel1-reader/blob/794d092a2e05f2492b223559c3c7b60903b2eff7/tests/download_data.py#L10), why not put this in a function `src/sentinel1_reader`?

vbrancat:
> @yunjunz's solution `wget` solution sounds good to me.
> 
> Since you've already written this [function](https://github.com/vbrancat/sentinel1-reader/blob/794d092a2e05f2492b223559c3c7b60903b2eff7/tests/download_data.py#L10), why not put this in a function `src/sentinel1_reader`? This would address #3.

I have a script (that I have never pushed) to download orbit ephemerides without credentials. Will address this in another PR.

yunjunz (approved):
Looks great!

hfattahi (approved):
LGTM!

PR #21, opened 2022-03-04
use shorter dir/file names like "s1reader"

yunjunz:
+ rename the repo directory and core module names to shorter ones, as discussed in https://github.com/opera-adt/s1-reader/discussions/14. 

+ update the corresponding usages due to the renaming


bin/s1_read.py
```diff
@@ -1,7 +1,7 @@
 import os
 import sys
 
-from sentinel1_reader import sentinel1_reader, sentinel1_orbit_reader
+from s1reader import reader, orbit
```
LiangJYu, line 4: I'm concerned renaming `sentinel1_reader, sentinel1_orbit_reader` to `reader, orbit` this too vague. @yunjunz What do you think about the following?
```suggestion
from s1reader import s1_reader, s1_orbit
```
 `s1_` is still much shorter than `sentinel1_` and it leaves no ambiguity as to purpose/origin.

yunjunz: yeah, that sounds a good idea.

yunjunz: I will adjust the naming in the PR accordingly to this.

yunjunz: Done.

LiangJYu (approved):
LGTM. I really like these changes.

vbrancat (approved):
LGTM. Thanks for these changes @yunjunz. Your nomenclature looks really coincise. We should make sure that none of the modules using the reader (e.g. COMPASS) will break after mergin this PR>

hfattahi (approved):
LGTM. Thanks Yunjun.

PR #23, opened 2022-03-23
return self as a dict

LiangJYu:
This PR adds ability return an object as a dict and will be in writing COMPASS runconfig to file.

TODO:
* ~~write out `'doppler', 'azimuth_fm_rate', 'border', 'orbit'`~~

### DIFF
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -386,7 +386,44 @@ class Sentinel1BurstSlc:
                 val = val.coords[0]
             elif isinstance(val, np.float64):
                 val = float(val)
-            elif key in ['doppler', 'azimuth_fm_rate', 'border', 'orbit']:
-                continue
+            elif key == 'azimuth_fm_rate':
+                temp = {}
+                temp['order'] = val.order
+                temp['mean'] = val.mean
+                temp['std'] = val.std
+                temp['coeffs'] = val.coeffs
+                val = temp
+            elif key == 'border':
+                val = self.border[0].wkt
+            elif key == 'doppler':
+                temp = {}
+
+                temp['poly1d'] = {}
+                temp['poly1d']['order'] = val.poly1d.order
+                temp['poly1d']['mean'] = val.poly1d.mean
+                temp['poly1d']['std'] = val.poly1d.std
+                temp['poly1d']['coeffs'] = val.poly1d.coeffs
+
+                temp['lut2d'] = {}
+                temp['lut2d']['x_start'] = val.lut2d.x_start
+                temp['lut2d']['x_spacing'] = val.lut2d.x_spacing
+                temp['lut2d']['y_start'] = val.lut2d.y_start
+                temp['lut2d']['y_spacing'] = val.lut2d.y_spacing
+                temp['lut2d']['length'] = val.lut2d.length
+                temp['lut2d']['width'] = val.lut2d.width
+                temp['lut2d']['data'] = val.lut2d.data
+
+                val = temp
+            elif key in ['orbit']:
+                temp = {}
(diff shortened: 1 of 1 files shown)


src/s1reader/s1_burst_slc.py
```diff
@@ -357,3 +357,25 @@ class represents a polynomial function of range
...
+        """
+        self_as_dict = {}
+        for key, val in self.__dict__.items():
+            if key == 'sensing_start':
+                val = str(val)
+            elif key == 'center':
+                val = val.coords[0]
+            elif isinstance(val, np.float64):
+                val = float(val)
+            elif key in ['doppler', 'azimuth_fm_rate', 'border', 'orbit']:
```
hfattahi, line 378: if you accept my suggestions above, then this line would change to :
```suggestion
            elif key in ['orbit']:
```

src/s1reader/s1_burst_slc.py
```diff
@@ -357,3 +357,25 @@ class represents a polynomial function of range
...
+           Dict representation as a dict
+        """
+        self_as_dict = {}
+        for key, val in self.__dict__.items():
+            if key == 'sensing_start':
+                val = str(val)
+            elif key == 'center':
+                val = val.coords[0]
+            elif isinstance(val, np.float64):
+                val = float(val)
```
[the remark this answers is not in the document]
hfattahi: and for azimuth fm rate which is a 1D poly we can do:

```
val = burst.azimuth_fm_rate.coeffs
```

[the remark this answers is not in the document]
hfattahi: For Doppler I don't think if we need to store both poly1D and lut2d. Let's just take the poly1d which is much simpler:

```
elif key=='doppler':
     val = burst.doppler.poly1d.coeffs
```

hfattahi (commented):
Thank you @LiangJYu . Here is few suggestions for border, Doppler and azimuth fm rate. If you accept those, then we are only left with orbit. Which we can discuss further.

vbrancat (approved):
Pending the suggestions proposed by @hfattahi LGTM

src/s1reader/s1_burst_slc.py
```diff
@@ -368,3 +368,62 @@ class represents a polynomial function of range
...
+                val = temp
+            elif key in ['orbit']:
+                temp = {}
+                temp['ref_epoch'] = str(val.reference_epoch)
+                temp['time'] = {}
+                temp['time']['first'] = val.time.first
+                temp['time']['spacing'] = val.time.spacing
+                temp['time']['last'] = val.time.last
+                temp['time']['size'] = val.time.size
+                temp['position'] = val.position.flatten().tolist()
```
hfattahi, line 425: Position is 2 dimensional with a shape of number_of_state_vectorx by 3. The three columns of the 2D array represent the coordinates in X, Y and Z directions. To avoid any confusion in reconstructing orbit from the stored metadata, I suggest that we be very clear on the three components: 
```suggestion
                temp['positionX'] = val.position[:,0].tolist()
                temp['positionY'] = val.position[:,1].tolist()
                temp['positionZ'] = val.position[:,2].tolist()
```

src/s1reader/s1_burst_slc.py
```diff
@@ -368,3 +368,62 @@ class represents a polynomial function of range
...
+                temp['velocity'] = val.velocity.flatten().tolist()
```
hfattahi, line 426: Similar to position, velocity also has three components let's avoid any confusion:
```suggestion
                temp['velocityX'] = val.velocity[:,0].tolist()
                temp['velocityY'] = val.velocity[:,1].tolist()
                temp['velocityZ'] = val.velocity[:,2].tolist()
                
```

Issue #22, opened 2022-03-23
read from SAFE files with and without measurements: 


hfattahi:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

Without this feature we won't be able to read unzipped files. Also we won't be able to read only downloaded metadata files. 

### Describe the feature request

s1reader currently only reads the ".zip" files. It should be able to read both zipfile and ".SAFE" files. For the latter we should be able to read a .SAFE file which may or may not contain any actual measurements. For example if we download only the annotation files, which is possible [using this](https://github.com/avalentino/asfsmd), we want to be able to read the downloaded metadata.

LiangJYu:
`s1_reader` [reads](https://github.com/opera-adt/s1-reader/blob/b8de577fe8c807f3375db1be7775032327cb6295/src/s1reader/s1_reader.py#L250) an XML from a provided path. The zip reader is a [wrapper](https://github.com/opera-adt/s1-reader/blob/b8de577fe8c807f3375db1be7775032327cb6295/src/s1reader/s1_reader.py#L451) around the XML reader.

hfattahi:
@LiangJYu yes I understand that. However that is not the interface we want at the high-level. Forexample the burst_from_xml requires the path to the xml and tiff files. User of the reader should not do that. burst_from_xml does not need to change. We can either add another function similar to burst_from_zip with the following arguments:

burst_from_safe(safe_path: str, orbit_path: str, swath_num: int, pol: str = 'vv')

or generalize burst_from_zip to ```get_burst(file_path: str, orbit_path: str, swath_num: int, pol: str = 'vv')``` where ```file_path``` can be a zip file or a unzipped version of it, i.e., ".SAFE" file. 

For the second need where we only have the annotation file, what we download is same as the zip file, with two differences. One it may not be zipped and it does not have measurements (i.e., tiff file).

PR #24, opened 2022-03-24
generalized SAFE and zip reader

LiangJYu:
This PR addresses #22 by adding a public-facing, generalized `load_bursts` method that:
1. loads from both zip and SAFE, 
2. doesn't require a tiff file to be found/included.

It is intended to replace the current public-facing `bursts_from_zip` method.

### DIFF
--- a/bin/s1_read.py
+++ b/bin/s1_read.py
@@ -6,9 +6,9 @@ import s1reader
 if __name__ == "__main__":
     '''testing script that prints burst info and write SLC to file'''
     # TODO replace with argparse
-    zip_path = sys.argv[1]
-    if not os.path.isfile(zip_path):
-        raise FileNotFoundError(f"{zip_path} does not exist")
+    path = sys.argv[1]
+    if not os.path.exists(path):
+        raise FileNotFoundError(f"{path} does not exist")
 
     i_subswath = int(sys.argv[2])
     if i_subswath < 1  or i_subswath > 3:
@@ -22,17 +22,19 @@ if __name__ == "__main__":
     orbit_dir = sys.argv[4]
     if not os.path.isdir(orbit_dir):
         raise NotADirectoryError(f"{orbit_dir} not found")
-    orbit_path = s1reader.get_orbit_file_from_dir(zip_path, orbit_dir)
+    orbit_path = s1reader.get_orbit_file_from_dir(path, orbit_dir)
 
-    bursts = s1reader.burst_from_zip(zip_path, orbit_path, i_subswath, pol)
+    bursts = s1reader.load_burst(path, orbit_path, i_subswath, pol)
 
     # print out IDs and lat/lon centers of all bursts
     for i, burst in enumerate(bursts):
         print(burst.burst_id, burst.center)
 
     # write to ENVI (default)
+    '''
     burst.slc_to_file('burst.slc')
     # write to geotiff
     burst.slc_to_file('burst.tif', 'GTiff')
     # write to VRT
     burst.slc_to_file('burst.vrt', 'VRT')
+    '''
--- a/bin/s1_read.py
+++ b/bin/s1_read.py
@@ -31,10 +31,8 @@ if __name__ == "__main__":
         print(burst.burst_id, burst.center)
 
     # write to ENVI (default)
-    '''
     burst.slc_to_file('burst.slc')
     # write to geotiff
     burst.slc_to_file('burst.tif', 'GTiff')
     # write to VRT
     burst.slc_to_file('burst.vrt', 'VRT')
-    '''
--- a/bin/s1_read.py
+++ b/bin/s1_read.py
@@ -24,7 +24,7 @@ if __name__ == "__main__":
         raise NotADirectoryError(f"{orbit_dir} not found")
     orbit_path = s1reader.get_orbit_file_from_dir(path, orbit_dir)
 
-    bursts = s1reader.load_burst(path, orbit_path, i_subswath, pol)
+    bursts = s1reader.load_bursts(path, orbit_path, i_subswath, pol)
 
     # print out IDs and lat/lon centers of all bursts
     for i, burst in enumerate(bursts):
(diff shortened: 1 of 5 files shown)


src/s1reader/__init__.py
```diff
@@ -1,4 +1,4 @@
 # top-level functions to be easily used
 from .s1_burst_slc import Sentinel1BurstSlc
-from .s1_reader import burst_from_zip, burst_from_xml
-from .s1_orbit import get_orbit_file_from_dir
\ No newline at end of file
+from .s1_reader import load_burst
```
hfattahi, line 3:
```suggestion
from .s1_reader import load_bursts
```

src/s1reader/s1_reader.py
```diff
@@ -437,16 +437,80 @@ def burst_from_zip(zip_path: str, orbit_path: str, swath_num: int, pol: str = 'v
...
+
+    # find annotation file
+    annotation_list = os.listdir(f'{safe_dir_path}/annotation')
+    f_annotation = [f for f in annotation_list if id_str in f]
+    if not f_annotation:
+        raise ValueError(f"burst {id_str} not in SAFE: {safe_dir_path}")
+    f_annotation = f'{safe_dir_path}/annotation/{f_annotation[0]}'
+
+    # find tiff file
+    measurement_list = os.listdir(f'{safe_dir_path}/measurement')
```
hfattahi, line 510: For the case that we may only download the annotations, we don't have measurement in the path. Therefore we would get error like this one:

```
Traceback (most recent call last):
  File "/u/aurora-r0/fattahi/burst_map/./test_read.py", line 14, in <module>
    bursts = s1reader.load_bursts(zip_path, orbit_path, swath_num, pol)
  File "/home/fattahi/tools/anaconda3/envs/burst_id/lib/python3.10/site-packages/s1reader/s1_reader.py", line 444, in load_bursts
    return _burst_from_safe_dir(path, id_str, orbit_path)
  File "/home/fattahi/tools/anaconda3/envs/burst_id/lib/python3.10/site-packages/s1reader/s1_reader.py", line 510, in _burst_from_safe_dir
    measurement_list = os.listdir(f'{safe_dir_path}/measurement')
FileNotFoundError: [Errno 2] No such file or directory: 'xmls/S1A_IW_SLC__1SDV_20190330T015159_20190330T015226_026561_02FA29_6720.SAFE/measurement'
```

hfattahi: Perhaps just check if `f'{safe_dir_path}/measurement'` exists and then go through the next few lines. If the measurement does not exist, let's throw a warning that measurement does not exist and proceed because we still can read the metadata.

hfattahi (commented):
Thanks Liang. Few minor comments which should be easy to fix and then this PR will be good to go.

src/s1reader/s1_reader.py
```diff
@@ -437,16 +438,85 @@ def burst_from_zip(zip_path: str, orbit_path: str, swath_num: int, pol: str = 'v
...
-            raise ValueError(f"polarization {pol} not in SAFE: {zip_path}")
+            raise ValueError(f"burst {id_str} not in SAFE: {zip_path}")
         f_annotation = f_annotation[0]
 
         # find tiff file
-        f_tiff = [f for f in z_file.namelist() if 'measurement' in f and id_str in f and 'tiff' in f][0]
-        f_tiff = f'/vsizip/{zip_path}/{f_tiff}'
+        f_tiff = [f for f in z_file_list
+                  if 'measurement' in f and id_str in f and 'tiff' in f]
+        f_tiff = f'/vsizip/{zip_path}/{f_tiff[0]}' if f_tiff else ''
```
hfattahi, line 481: if f_tiff is empty, let's let the user know. Same warning as you added at line 517.

hfattahi (approved):
Thanks Liang for the quick fix. Left one more comment to add a warning for burst_from_zip when the input does not have measurement. 
Also would you please check if "tiff" exists [here](https://github.com/opera-adt/s1-reader/blob/main/src/s1reader/s1_burst_slc.py#L273) and if it does not exist throw a meaningful message. e.g., "The input file does not contain image data. It only contains metadata. "

bin/s1_read.py
```diff
@@ -22,17 +22,19 @@
...
 
-    bursts = s1reader.burst_from_zip(zip_path, orbit_path, i_subswath, pol)
+    bursts = s1reader.load_bursts(path, orbit_path, i_subswath, pol)
 
     # print out IDs and lat/lon centers of all bursts
     for i, burst in enumerate(bursts):
         print(burst.burst_id, burst.center)
 
     # write to ENVI (default)
+    '''
```
vbrancat, line 34: May be remove this commented parts if not needed

src/s1reader/s1_reader.py
```diff
@@ -437,16 +438,85 @@ def burst_from_zip(zip_path: str, orbit_path: str, swath_num: int, pol: str = 'v
...
+    '''
+
+    # find annotation file
+    annotation_list = os.listdir(f'{safe_dir_path}/annotation')
+    f_annotation = [f for f in annotation_list if id_str in f]
+    if not f_annotation:
+        raise ValueError(f"burst {id_str} not in SAFE: {safe_dir_path}")
+    f_annotation = f'{safe_dir_path}/annotation/{f_annotation[0]}'
+
+    # find tiff file if measurement directory found
```
vbrancat, line 510: In the case we get only the XML annotation file, there is no "measurements" . Maybe, in that case, we should not check it?

vbrancat (approved):
Thanks a lot for the quick fix @lyu. It looks good to me. My only concern is related to checking the `measurements` directory/path when we get assigned only with the XML annotation file. We should check that when we get the XML, we extract data from it and don't required to look at the `measurement `dir/path

yunjunz (commented):
I like this PR a lot, it really simplifies the usage on the user side. Besides the more meaningful warning msg that Heresh has pointed out, the rest looks good to me.

src/s1reader/s1_reader.py
yunjunz: @vbrancat I don't quite get your comment here: we don't know whether there are measurements or not in advance, that's why the code checks it here, isn't it?

vbrancat: My bad!

yunjunz: Okay, it's good to make sure that we are on the same page of the code.

PR #26, opened 2022-03-30
README: add license and pre-alpha warning

yunjunz:
+ README: 
   - add license info, to be consistent with other opera-adt repos.
   - and pre-alpha warning msg.

+ requirements:
   - add `python>=3.9` to be consistent with `setup.cfg`
   - comment out `isce3` as conda installed version does not work yet.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
+python>=3.9
 numpy
 gdal
-isce3
+#isce3  #comment out temporarily as conda installed version does not work
 shapely
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 python>=3.9
 numpy
 gdal
-#isce3    # [temporary] install isce3 from stratch instead, to stay in sync with isce3 development.
+#isce3    # since the conda-installed isce3 is not the most updated version, installing isce3 from stratch is recommended, to stay in sync with isce3 development.
 shapely
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 python>=3.9
 numpy
 gdal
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,4 +1,5 @@
+python>=3.9
 numpy
 gdal
-isce3
+#isce3  #comment out temporarily as conda installed version does not work
```
vbrancat, line 4: What do you mean with "it does not work"? Can you be more specific on this?

yunjunz: I actually don't know more specific details. I learned that the conda installed version does not work from our last meeting, I thought that was the reason I was recommended to use build isce3 from the source, isn't it?

vbrancat: The version of ISCE3 obtained through `conda` is not the most updated one. This means that it does not contain the most recent development pushed to the GitHub repo (@rtburns-jpl can obviously correct me on that). To stay in sync with ISCE3 development, it is recommended to install ISCE3 from scratch. For basic functionalities and if GPU processing is not required, the `conda` version should be sufficient.

yunjunz: Thank you @vbrancat, that makes all the sense. I updated the comment accordingly. Since we are mainly users currently, I think it is still a good idea to comment this out in the requirements.txt file, for now, to facilitate our own installation process.

vbrancat: I am fine to leave it as commented out (as it should be if people want to install it from scratch) bu modify the comment to reflect that is not the most updated version to work in combination to the ADT repos. Merci

vbrancat (approved):
LGTM pending the comment on the `isce3` conda package.

PR #29, opened 2022-04-27
Add handy properties

LiangJYu:
This PR adds the following handy properties to `Sentinel1BurstSlc` class:
* sensing mid and stop as properties to be consistent with sensing start.
* length and width broken out of shape attribute

### DIFF
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -386,7 +386,44 @@ class Sentinel1BurstSlc:
                 val = val.coords[0]
             elif isinstance(val, np.float64):
                 val = float(val)
-            elif key in ['doppler', 'azimuth_fm_rate', 'border', 'orbit']:
-                continue
+            elif key == 'azimuth_fm_rate':
+                temp = {}
+                temp['order'] = val.order
+                temp['mean'] = val.mean
+                temp['std'] = val.std
+                temp['coeffs'] = val.coeffs
+                val = temp
+            elif key == 'border':
+                val = self.border[0].wkt
+            elif key == 'doppler':
+                temp = {}
+
+                temp['poly1d'] = {}
+                temp['poly1d']['order'] = val.poly1d.order
+                temp['poly1d']['mean'] = val.poly1d.mean
+                temp['poly1d']['std'] = val.poly1d.std
+                temp['poly1d']['coeffs'] = val.poly1d.coeffs
+
+                temp['lut2d'] = {}
+                temp['lut2d']['x_start'] = val.lut2d.x_start
+                temp['lut2d']['x_spacing'] = val.lut2d.x_spacing
+                temp['lut2d']['y_start'] = val.lut2d.y_start
+                temp['lut2d']['y_spacing'] = val.lut2d.y_spacing
+                temp['lut2d']['length'] = val.lut2d.length
+                temp['lut2d']['width'] = val.lut2d.width
+                temp['lut2d']['data'] = val.lut2d.data
+
+                val = temp
+            elif key in ['orbit']:
+                temp = {}
+                temp['ref_epoch'] = val.reference_epoch
+                temp['time'] = {}
+                temp['time']['first'] = val.time.first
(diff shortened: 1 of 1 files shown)


src/s1reader/s1_burst_slc.py
```diff
@@ -313,6 +313,17 @@ def get_sensing_mid(self):
...
 
+    def get_sensing_end(self):
+        '''Returns sensing end as datetime.datetime object.
+
+        Returns:
+        --------
+        _ : datetime.datetime
+            Sensing end as datetime.datetime object.
+        '''
+        d_seconds = (self.shape[0] - 1) * self.azimuth_time_interval
```
hfattahi, line 324: Now that you are adding handy parameters, another handy parameter is length and width. :)

hfattahi (approved):

src/s1reader/s1_burst_slc.py
```diff
@@ -313,6 +313,17 @@ def get_sensing_mid(self):
         d_seconds = 0.5 * (self.shape[0] - 1) * self.azimuth_time_interval
         return self.sensing_start + datetime.timedelta(seconds=d_seconds)
 
+    def get_sensing_end(self):
```
yunjunz, line 316: Should we call this `get_sensing_stop()`? It matches with `sensing_start`..

src/s1reader/s1_burst_slc.py
```diff
@@ -431,3 +420,35 @@ def as_dict(self):
...
+    @property
+    def sensing_stop(self):
+        '''Returns sensing end as datetime.datetime object.
+
+        Returns:
+        --------
+        _ : datetime.datetime
+            Sensing end as datetime.datetime object.
+        '''
+        d_seconds = (self.shape[0] - 1) * self.azimuth_time_interval
```
yunjunz, line 445: Could you help understand why there is `- 1` here for both sensing_stop and sensing_mid?

LiangJYu: The `- 1` accounts for the first element. Without the `- 1`, the `sensing_stop` returns a time value outside radar grid.

hfattahi: makes sense for stop but not for mid.

src/s1reader/s1_burst_slc.py
```diff
@@ -431,3 +420,35 @@ def as_dict(self):
...
+    @property
+    def sensing_mid(self):
+        '''Returns sensing mid as datetime.datetime object.
+
+        Returns:
+        --------
+        _ : datetime.datetime
+            Sensing mid as datetime.datetime object.
+        '''
+        d_seconds = 0.5 * (self.shape[0] - 1) * self.azimuth_time_interval
```
hfattahi:
```suggestion
        d_seconds = 0.5 * self.shape[0] * self.azimuth_time_interval
```

src/s1reader/s1_burst_slc.py
yunjunz: For `sensing_stop`, I feel like this is similar to "a pixel is an area or a point" topic, if `sensing_start` is the beginning of the first pixel and `sensing_stop` is the end of the last pixel, then we should not use `- 1`, am I understanding it right?

@LiangJYu could you maybe show the value of the sensing_stop and radar grid boundaries here for us to better understand?

LiangJYu: image
[compute sensing stop.xlsx](https://github.com/opera-adt/s1-reader/files/8575769/compute.sensing.stop.xlsx)

hfattahi: @yunjunz sensing_stop here is actually the time-tag of the last sample. This is just the conversion between index and time. If sensing_start represents the beginning of the first pixel, sensing stop represents the beginning of the last pixel.

yunjunz: Thank you @LiangJYu and @hfattahi for the explanation.

yunjunz (approved):
Looks great to me.

PR #33, opened 2022-05-02
add burst metadata: burst_duration & swath_name

yunjunz:
This PR adds two more metadata to the burst class.

+ s1_slc_burst: add two more metadata:
   - burst_duration
   - swath_name

+ README: update usage to load_bursts()

### DIFF
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -448,9 +448,26 @@ class Sentinel1BurstSlc:
         return self.sensing_start + datetime.timedelta(seconds=d_seconds)
 
     @property
+    def burst_duration(self):
+        '''Returns burst sensing duration as float in seconds.
+
+        Returns:
+        --------
+        _ : float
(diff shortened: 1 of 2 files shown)


src/s1reader/s1_burst_slc.py
```diff
@@ -447,10 +447,27 @@ def sensing_stop(self):
...
+    @property
+    def burst_duration(self):
+        '''Returns burst sensing duration as float in seconds.
+
+        Returns:
+        --------
+        _ : float
+            Burst sensing duration as float in seconds.
+        '''
+        return self.azimuth_time_interval * self.shape[0]
```
LiangJYu, line 459:
```suggestion
        return self.azimuth_time_interval * self.length
```
Not 100% on the following: I think a name is clearer than indexing a tuple.

LiangJYu (approved):
LGTM

hfattahi (approved):
LGTM!

vbrancat (approved):
LGTM.

PR #34, opened 2022-05-03
Doppler shift

hfattahi:
This PR add functionality to compute range delay caused by Doppler shift as described by Gisinger et al 2021.
Note that to be able to reuse some the Carrier computation function that we had, I had to break it to two functions, one to compute Kt and one that computes the carrier and converts to phase in radian.

### DIFF
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -458,13 +458,13 @@ class Sentinel1BurstSlc:
 
         Returns
         -------
-        1D array: int
-           The index of samples in range direction
-        1D array: int
-           The index of samples in azimuth direction
-        2D array: float
+        x : int
+           The index of samples in range direction as an 1D array
+        y : int
+           The index of samples in azimuth direction as an 1D array
+        total_doppler : float
            Total Doppler which is the sum of the geometrical Doppler and
-           beam steering induced Doppler [Hz]
+           beam steering induced Doppler [Hz] as a 2D array
         """
 
         x = np.arange(0, self.width, xstep, dtype=int)
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -111,11 +111,11 @@ class AzimuthCarrierComponents:
     eta_ref: float
 
     @property
-    antenna_steering_doppler(self):
+    def antenna_steering_doppler(self):
         return self.kt * (self.eta - self.eta_ref)
 
     @property
-    carrier(self):
+    def carrier(self):
         return np.pi * self.kt * ((self.eta - self.eta_ref) ** 2)
 
 @dataclass(frozen=True)
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -195,7 +195,8 @@ class Sentinel1BurstSlc:
     # window parameters
     range_window_type: str
     range_window_coefficient: float
-
+    rank: int # The number of PRI between transmitted pulse and return echo.
+    prf_raw_data: float  # Pulse repetition frequency (PRF) of the raw data [Hz]
 
     def as_isce3_radargrid(self):
         '''Init and return isce3.product.RadarGridParameters.
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -27,8 +27,13 @@ def compute_az_carrier_frequency(burst, orbit, offset, position):
 
     Returns
     -------
-    carr: np.ndarray
-       Azimuth carrier
+    eta: zero-Doppler azimuth time centered in the middle of the burst
+    eta_ref: refernce time
+    Kt: Doppler centroid rate in the focused TOPS SLC data [Hz/s]
+
+    Reference
+    ---------
+    https://sentinels.copernicus.eu/documents/247904/0/Sentinel-1-TOPS-SLC_Deramping/b041f20f-e820-46b7-a3ed-af36b8eb7fa0
     '''
 
     # Get burst sensing mid relative to orbit reference epoch
@@ -59,7 +64,25 @@ def compute_az_carrier_frequency(burst, orbit, offset, position):
     return kt, eta, eta_ref
 
 def compute_az_carrier(burst, orbit, offset, position):
-    
+    '''
+    Estimate azimuth carrier and store in numpy arrary
+
+    Parameters
+    ----------
+    burst: Sentinel1BurstSlc
+       Sentinel1 burst object
+    orbit: isce3.core.Orbit
+       Sentinel1 orbit ephemerides
+    offset: float
+       Offset between reference and secondary burst
+    position: tuple
+       Tuple of locations along y and x directions
+
+    Returns
+    -------
+    carr: np.ndarray
+       Azimuth carrier
+    '''
     kt, eta, eta_ref = compute_az_carrier_frequency(burst, orbit, offset, position)
     carr = np.pi * kt * ((eta - eta_ref) ** 2)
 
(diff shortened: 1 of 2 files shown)


src/s1reader/s1_burst_slc.py
```diff
@@ -360,6 +389,68 @@ class represents a polynomial function of range
...
 
+    def total_doppler(self, xstep=500, ystep=50):
+        """
+        Compute total Doppler which is the sum of two components: 
+        (1) the geometrical Doppler induced by the relative movement 
+        of the sensor and target
+        (2) the TOPS specicifc Doppler caused by the electric steering 
+        of the beam along the azimuth direction resulting in Doppler varying
+        with azimuth time. 
+        Parameters
```
vbrancat, line 400: nit: space

src/s1reader/s1_burst_slc.py
```diff
@@ -360,6 +389,68 @@ class represents a polynomial function of range
...
+        ----------
+        xstep: int
+            Spacing along x direction [pixels]
+        ystep: int
+            Spacing along y direction [pixels]
+
+        Returns
```
vbrancat, line 407: Modify docstring to reflect returned variables (not only total doppler but also `x,` `y` vectors)

src/s1reader/s1_burst_slc.py
```diff
@@ -360,6 +389,68 @@ class represents a polynomial function of range
...
+        antenna_steering_Doppler = kt*(eta - eta_ref)
+
+        slant_range = self.starting_range + x * self.range_pixel_spacing
+        geometrical_doppler = self.doppler.poly1d.eval(slant_range)
+
+        total_Doppler = antenna_steering_Doppler + geometrical_doppler
+
+        return x, y, total_Doppler
+
+    def range_delay_caused_by_doppler_shift(self, xstep=500, ystep=50):
```
vbrancat, line 429: may be `doppler_shift_range_delay`?

src/s1reader/s1_burst_slc.py
```diff
@@ -360,6 +389,68 @@ class represents a polynomial function of range
 
         return az_carrier_poly
 
+    def total_doppler(self, xstep=500, ystep=50):
```
vbrancat, line 392: I am not a fun of `total_doppler`. Maybe we should be more explicit with the name of the function and call it `geometry_steering_doppler`. I know the doc string give an explanation but the name of the function sounds a bit confusing.

src/s1reader/s1_burst_slc.py
```diff
@@ -360,6 +389,68 @@ class represents a polynomial function of range
...
+        kt, eta, eta_ref = compute_az_carrier_frequency(self, self.orbit,
+                                        offset=0.0,
+                                        position=(y_mesh, x_mesh))
+
+        antenna_steering_Doppler = kt*(eta - eta_ref)
+
+        slant_range = self.starting_range + x * self.range_pixel_spacing
+        geometrical_doppler = self.doppler.poly1d.eval(slant_range)
+
+        total_Doppler = antenna_steering_Doppler + geometrical_doppler
```
vbrancat, line 425: Having capital letters is not conforming to pep8 notation (which we have been following in the Sentinel reader)

vbrancat (commented):
Thanks for this PR, @hfattahi. The math looks good to me from what I remember from Gisinger paper. I just have a couple of minor style comments which should not take too much time to address (I can address them if you want)

yunjunz:
Here is a comparison of 3 subswaths, between the S1-ETAD product and this PR: they are very close.

Screen Shot 2022-05-05 at 3 27 45 PM

hfattahi:
> Here is a comparison of 3 subswaths, between the S1-ETAD product and this PR: they are very close.
> 
> Screen Shot 2022-05-05 at 3 27 45 PM

Thank you @yunjunz . I think the differences are in the order of 1e-11 or smaller which is equivalent to 1.5 mm or smaller. This is totally negligible.

src/s1reader/s1_burst_slc.py
hfattahi: I was not very proud of the name too. But is "doppler_shift_range_delay" clear on "range delay caused by Doppler shift" and not confused by "Doppler shift caused by range delay"? 
I will change to your suggestion for now but let me know if you or yunjun find a better name!

yunjunz: how about `doppler_induced_range_shift()`, which is basically the subsection title in the ETAD [ATBD page 44](https://sentinels.copernicus.eu/documents/247904/4629150/Sentinel-1-ETAD-Algorithm-Technical-Baseline-Document.pdf) ?

vbrancat: @yunjunz I like that :D

src/s1reader/s1_burst_slc.py
yunjunz: ETAD ATBD uses the term "Doppler centroid frequency of focused data", so maybe `focused_doppler()`?

yunjunz:
Could we also move the two new functions next to the `bistatic_delay()`?

src/s1reader/s1_burst_slc.py
hfattahi: Thanks Yunjun and Virginia. focused_doppler is technically correct and represent the total_doppler. However, I think ```geometry_steering_doppler``` is a bit more clear and tops specific. It may sound like one component of Doppler which is due to geometricaly steering, so I added ```and```, i.e., I went with ```geometry_and_steering_doppler``` if ok with you both.

yunjunz (approved):
LGTM!

hfattahi:
@vbrancat @LiangJYu Could you take another look at this PR. I think the comments have been addressed. Yunjun's test confirms that the implementation is consistent with ETAD.

src/s1reader/s1_burst_slc.py
```diff
@@ -10,7 +10,7 @@
 
 
 # Other functionalities
-def compute_az_carrier(burst, orbit, offset, position):
+def compute_az_carrier_frequency(burst, orbit, offset, position):
```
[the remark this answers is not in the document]
hfattahi: I agree on moving the function to the class method. I was actually going to suggest the same thing after this PR. 
If ok with you, I would prefer to keep the more clear names of "compute_az_carrier_frequency". Perhaps we should shorten to "azimuth_carrier_frequency" though.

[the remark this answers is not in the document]
LiangJYu: `deramp_components` is ambiguous. I suggested it as an alternative to `compute_az_carrier_frequency` since none of the return 3 values are `az_carrier_frequency`.

What do you think about not splitting `compute_az_carrier_frequency` into 2 functions and instead adding `kt`, `eta`, `eta_ref` as additional return values?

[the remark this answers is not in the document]
hfattahi: I think it is very useful to split the function. Because Kt has multiple usage and only one of them is to compute the phase ramp.

LiangJYu (approved):
LGTM

src/s1reader/s1_burst_slc.py
```diff
@@ -155,6 +104,19 @@ class represents a polynomial function of range
...
+    kt: np.ndarray
+    eta: float
+    eta_ref: float
+
+    @property
+    antenna_steering_doppler(self):
+        return self.kt * (self.eta - self.eta_ref)
+
+    @property
+    carrier(self):
```
LiangJYu, line 118:
```suggestion
    def antenna_steering_doppler(self):
        return self.kt * (self.eta - self.eta_ref)

    @property
    def carrier(self):
```
Missing `def`.

LiangJYu:
I have successfully tested the most recent fixes from this branch with COMPASS main . Can someone else please confirm? Thanks.

hfattahi:
@yunjunz would you please redo the test above you did to make sure everything still looks good in term of comparing the estimated shift vs ETAD?

yunjunz:
The difference between the two in terms of min/max varies from 2e-12 and 2e-11 seconds, so it's still good to go!

hfattahi:
Thank you @LiangJYu for your improvements to this PR. Thanks @yunjunz for the very useful tests.

PR #37, opened 2022-05-11
retrieve bursts by ID

LiangJYu:
This PR adds ability to filter bursts from `s1_reader.load_bursts` by ID by adding a list of burst IDs to filter against.

```python
# returns all bursts in given SAFE zip/directory
s1reader.load_bursts(safe_zip_path, orbit_path, i_subswath, pol)

# if provided IDs exist, returns only bursts listed
s1reader.load_bursts(safe_dir_path, orbit_path, i_subswath, pol, ['t71_iw1_b848', 't71_iw1_b849'])
```

### DIFF
--- a/src/s1reader/s1_reader.py
+++ b/src/s1reader/s1_reader.py
@@ -425,7 +425,8 @@ def _is_zip_annotation_xml(path: str, id_str: str) -> bool:
         return True
     return False
 
-def load_bursts(path: str, orbit_path: str, swath_num: int, pol: str = 'vv'):
+def load_bursts(path: str, orbit_path: str, swath_num: int, pol: str = 'vv',
+                burst_ids: list[str] = []):
     '''Find bursts in a Sentinel-1 zip file or a SAFE structured directory.
 
     Parameters:
@@ -438,13 +439,14 @@ def load_bursts(path: str, orbit_path: str, swath_num: int, pol: str = 'vv'):
         Integer of subswath of desired burst. {1, 2, 3}
     pol : str
         Polarization of desired burst. {hh, vv, hv, vh}
+    burst_ids : list[str]
+        List of burst IDs to be returned. Default of None returns all bursts.
 
     Returns:
     --------
     bursts : list
         List of Sentinel1BurstSlc objects found in annotation XML.
     '''
-
     if swath_num < 1 or swath_num > 3:
         raise ValueError("swath_num not <1 or >3")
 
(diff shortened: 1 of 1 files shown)


src/s1reader/s1_reader.py
```diff
@@ -438,13 +439,14 @@ def load_bursts(path: str, orbit_path: str, swath_num: int, pol: str = 'vv'):
         Integer of subswath of desired burst. {1, 2, 3}
     pol : str
         Polarization of desired burst. {hh, vv, hv, vh}
+    burst_ids : list[str]
+        List of burst IDs to be returned. Default of None returns all bursts.
```
vbrancat, line 443: Docstring is a bit confusing. It is the `Sentinel1BurstSlc` object for the specified burst ID that is returned (not really the burst ID XD)

src/s1reader/s1_reader.py
```diff
@@ -425,7 +425,8 @@ def _is_zip_annotation_xml(path: str, id_str: str) -> bool:
         return True
     return False
 
-def load_bursts(path: str, orbit_path: str, swath_num: int, pol: str = 'vv'):
+def load_bursts(path: str, orbit_path: str, swath_num: int, pol: str = 'vv',
```
vbrancat, line 428:
```suggestion
def load_bursts(path: str, orbit_path: str, swath_num: int, pol: str='vv',
```

 I think within the brackets, spaces between operators can be removed

vbrancat (approved):
LGTM. No major comments. Make sure to address the codacy issues before merging

src/s1reader/s1_reader.py
hfattahi: I agree. How about: "List of burst IDs for which their Sentinel1BurstSlc objects will be returned. Default of None returns all bursts."

src/s1reader/s1_reader.py
```diff
@@ -459,12 +461,22 @@ def load_bursts(path: str, orbit_path: str, swath_num: int, pol: str = 'vv'):
...
+        bursts = _burst_from_zip(path, id_str, orbit_path)
     else:
         raise ValueError(f'{path} is unsupported')
 
+    # ensure burst IDs is a list
+    if not isinstance(burst_ids, list) and isinstance(burst_ids, str):
+        burst_ids = [burst_ids]
+
+    if burst_ids:
+        return [b for b in bursts if b.burst_id in burst_ids]
```
hfattahi, line 475: I wonder if we should provide some useful informations here before returning. There would be cases that all the requested burst_ids are found in the zip file and returned. Also there will be situations that only part of the requested burst_ids are found in the zip file or maybe even None of them found. I think it will be very helpful to print/log some useful information on which burst_id is found and which one was not found before returning the list.

yunjunz (approved):
Besides the comments from Virginia and Heresh, the rest looks good to me.

src/s1reader/s1_reader.py
```diff
@@ -467,14 +477,18 @@ def load_bursts(path: str, orbit_path: str, swath_num: int, pol: str = 'vv',
...
-        return bursts
+        bursts = [b for b in bursts if b.burst_id in burst_ids]
+
+        burst_ids_found = set([b.burst_id for b in bursts])
+
+        if not burst_ids_found:
+            warnings.warn("None of provided bursts IDs found")
+
+        if burst_ids_found != set(burst_ids):
+            warnings.warn("Not all provided burst IDs found")
```
hfattahi, line 489: It will be really helpful to know which one was not found.

src/s1reader/s1_reader.py
```diff
@@ -485,8 +485,10 @@ def load_bursts(path: str, orbit_path: str, swath_num: int, pol: str='vv',
         if not burst_ids_found:
             warnings.warn("None of provided bursts IDs found")
 
-        if burst_ids_found != set(burst_ids):
-            warnings.warn("Not all provided burst IDs found")
+        set_burst_ids = set(burst_ids)
+        if burst_ids_found != set_burst_ids:
+            diff = set_burst_ids.difference(burst_ids_found)
+            warnings.warn(f'Following burst IDs found: {diff}')
```
hfattahi, line 491: I think it is giving the bursts which were not found. Can you recheck?
```suggestion
            warnings.warn(f'Following burst IDs were NOT found: {diff}')
```
It also does not harm to say which burst IDs were found.

hfattahi (approved):
Thank @LiangJYu. LGTM!

PR #38, opened 2022-05-16
unit tests

LiangJYu:
This PR adds unit tests to the 3 modules in s1-reader. The tests are currently very sparse and serve as a frame to build on.

**Usage:**
1. checkout `unit_test` branch from my fork
2. cd to `tests` directory
3. in terminal call: `pytest`

**To do:**
* test and validate more attributes


requirements.yaml
```diff
@@ -0,0 +1,9 @@
+name: test
```
yunjunz, line 1: This file is commonly named `environment.yaml`, shall we use that here too?

Maybe name this conda environment as `s1reader` to be distinguishable with future tests for other packages like `compass`?

yunjunz:
Thank you @LiangJYu, following your note above, the test runs nicely.

How did you prepare the SAFE zip file? I notice there are measurement folder and tiff files but with 0 in size.

LiangJYu:
> Thank you @LiangJYu, following your note above, the test runs nicely.
> 
> How did you prepare the SAFE zip file? I notice there are measurement folder and tiff files but with 0 in size. Maybe write this in a README.md file within the tests folder?

The 0 size files were `touch`d with a correct name and path. I will add a README to tests folder.

tests/test_bursts.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+from s1reader.s1_burst_slc import Sentinel1BurstSlc
+
+def test_burst(bursts):
+    for i, burst in enumerate(bursts):
+        expected_burst_id = f't71_iw3_b{844 + i}'
+        assert burst.burst_id == expected_burst_id
```
hfattahi, line 8: Thanks @LiangJYu . This is great start. I think we should go ahead and start checking all different burst data members for the specific burst of interest. You need to simply lookup the xml file manually and compare the extracted value by the reader with what you expected. 

Let's start with scalars and simpler parameters. Then let's discuss for more complicated parameters or methods (like azimuth_carrier etc).

.github/workflows/unit_tests.yaml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v2
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          activate-environment: test
+          auto-update-conda: true
+          environment-file: requirements.yaml
```
yunjunz, line 16:
```suggestion
          environment-file: environment.yaml
```

.github/workflows/unit_tests.yaml
```diff
@@ -0,0 +1,25 @@
...
+  unit_tests:
+    name: Unit tests
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v2
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          activate-environment: test
```
yunjunz, line 14:
```suggestion
          activate-environment: s1reader
```

.github/workflows/unit_tests.yaml
```diff
@@ -0,0 +1,25 @@
...
+          auto-update-conda: true
+          environment-file: requirements.yaml
+          python-version: 3.9
+          auto-activate-base: false
+      - name: Run pytest
+        shell: bash -l {0}
+        run: |
+          conda activate test
```
yunjunz, line 22:
```suggestion
          conda activate s1reader
```

environment.yaml
```diff
@@ -0,0 +1,9 @@
+name: test
```
yunjunz, line 1:
```suggestion
name: s1reader
```

LiangJYu:
Thus far, the checks for simple attributes have been implemented. I'm unsure how to test the following more complex attributes.

orbit: Ensure all state vectors within sensing start and stop? Check if position actually above surface of the earth? Check if velocity fast enough?

poly1D: These vary with burst so the verbose method would be to store and check order, std, mean, and coeffs of each. I can't think of a cleaner way.

lut2D: Can this be skiped since this is derived from poly1d?

hfattahi:
@LiangJYu  for the lut2d you can check if we are able to eval successfully compared to poly1d. For example :
```
np.isclose(burst.doppler.lut2d.eval(t0.total_seconds(),r0), burst.doppler.poly1d.eval(r0))
```
t0 and r0 can be the same middle of swath in my previous comment.

LiangJYu:
@hfattahi @yunjunz @vbrancat With the new tests added, I think this PR is ready for another look. Thanks!

hfattahi (approved):
LGTM! Thanks @LiangJYu for addressing my comments.

seongsujeong:
Hi @LiangJYu @yunjunz I am about to merge the annotation reader in PR #48. The new implementation loads IPF version from `manifest.safe` file in SAFE .zip file, which is currently not included in the [test data](https://github.com/opera-adt/s1-reader/blob/main/tests/data/S1A_IW_SLC__1SDV_20200511T135117_20200511T135144_032518_03C421_7768.zip). It would be great if the test data is updated with manifest.safe file included, so that the new PR can pass CircieCI.

Issue #40, opened 2022-05-18
Orbit from header


hfattahi:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

Currently the reader only works if the external orbit file is provided. 

### Describe the feature request

Since the annotation file already contains the less precise orbit, I think the default behavior should be to extract the orbit from external orbit if the orbit file is provided. If the orbit file is not provided let's just extract the orbit from the header (annotation). 

In the annotation file, the orbits are in <generalAnnotation/orbitList> . [isce2](https://github.com/isce-framework/isce2/blob/main/components/isceobj/Sensor/TOPS/Sentinel1.py#L794) implementation can be reused.

PR #41, opened 2022-05-18
WIP ability to load orbit from annotation XML

LiangJYu:
This PR address #40 and add:
* ability to load orbit from annotation XML
* moved orbit file reader to s1_orbit
* default to annotation orbit

To do:
* more testing
* update documentation

### DIFF
--- a/src/s1reader/s1_orbit.py
+++ b/src/s1reader/s1_orbit.py
@@ -1,5 +1,8 @@
 import datetime
 import os
+import xml.etree.ElementTree as ET
+
+import isce3
 
 # date format used in file names
 FMT = "%Y%m%dT%H%M%S"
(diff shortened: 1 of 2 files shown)


hfattahi:
Thank you @LiangJYu , It looks good. I gave it a try and for one case I got the following error message from our isce3 orbit constructor:
```
ValueError: non-uniform spacing between state vectors encountered - interval between state vector at position 1 and state vector at position 2 is 9.999999 s, expected 10.000001 s
```
complaining that the orbit is not uniform. We know that we have a very strict threshold in isce3 about orbit uniformity. Let's discuss if we should allow users to relax that threshold to be able to load some not very precise orbits at their own risk.

LiangJYu:
> Thank you @LiangJYu , It looks good. I gave it a try and for one case I got the following error message from our isce3 orbit constructor:
> 
> ```
> ValueError: non-uniform spacing between state vectors encountered - interval between state vector at position 1 and state vector at position 2 is 9.999999 s, expected 10.000001 s
> ```
> 
> complaining that the orbit is not uniform. We know that we have a very strict threshold in isce3 about checking the orbit uniform sampling. Let's discuss if we should allow users to relax that threshold to be able to load some not very precise orbits at their own risk.

@hfattahi what do you think about modifying the ISCE3 [orbit constructors](https://github.com/isce-framework/isce3/blob/49f18304a8fcba33dbf33347a64b7f89117cdcf6/cxx/isce3/core/Orbit.cpp#L20) to take a relaxed threshold? It doesn't appear too complicated to do in ISCE3. Here's what I think needs to be done:
* Utilize the [`iscClose()`](https://github.com/isce-framework/isce3/blob/49f18304a8fcba33dbf33347a64b7f89117cdcf6/cxx/isce3/core/DateTime.h#L99=) that is overload to accepts an user specified threshold.
* Modify the `Orbit` constructor and [`getOrbitTime`](https://github.com/isce-framework/isce3/blob/49f18304a8fcba33dbf33347a64b7f89117cdcf6/cxx/isce3/core/detail/BuildOrbit.cpp#L31=) to accept a user threshold while defaulting to current threshold

hfattahi:
> > Thank you @LiangJYu , It looks good. I gave it a try and for one case I got the following error message from our isce3 orbit constructor:
> > ```
> > ValueError: non-uniform spacing between state vectors encountered - interval between state vector at position 1 and state vector at position 2 is 9.999999 s, expected 10.000001 s
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > complaining that the orbit is not uniform. We know that we have a very strict threshold in isce3 about checking the orbit uniform sampling. Let's discuss if we should allow users to relax that threshold to be able to load some not very precise orbits at their own risk.
> 
> @hfattahi what do you think about modifying the ISCE3 [orbit constructors](https://github.com/isce-framework/isce3/blob/49f18304a8fcba33dbf33347a64b7f89117cdcf6/cxx/isce3/core/Orbit.cpp#L20) to take a relaxed threshold? It doesn't appear too complicated to do in ISCE3. Here's what I think needs to be done:
> 
> * Utilize the [`iscClose()`](https://github.com/isce-framework/isce3/blob/49f18304a8fcba33dbf33347a64b7f89117cdcf6/cxx/isce3/core/DateTime.h#L99=) that is overload to accepts an user specified threshold.
> * Modify the `Orbit` constructor and [`getOrbitTime`](https://github.com/isce-framework/isce3/blob/49f18304a8fcba33dbf33347a64b7f89117cdcf6/cxx/isce3/core/detail/BuildOrbit.cpp#L31=) to accept a user threshold while defaulting to current threshold

There can be two approaches to solve this issue I guess. One would be to allow irregular state vector sampling and let the orbit interpolator handle things. I believe we already have an open issue in isce3 repo and it will eventually come in. 

The other solution is what you suggested. It is simpler but a bit risky. However, the users of these APIs are really advanced developers and they should know what they are doing if they decide to relax the threshold. I'm ok if we allow for a threshold to be passed to the orbit constructor. By default it won't be needed and would use the current strict threshold. Let's discuss on isce3 repo.

Issue #43, opened 2022-05-20
[Bug]: first_valid_line ValueError: True not in list


otirmizi:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

```
Traceback (most recent call last):
  File "/DataFS/otirmizi/JPL/asc_stack/s1_geo_stack.py", line 507, in <module>
    main(args.slc_dir, args.dem_file, args.burst_id, args.ref_date,
  File "/DataFS/otirmizi/JPL/asc_stack/s1_geo_stack.py", line 484, in main
    bursts = load_bursts(safe, orbit_path, subswath)
  File "/home/users/otirmizi/tools/s1-reader/src/s1reader/s1_reader.py", line 476, in load_bursts
    bursts = _burst_from_zip(path, id_str, orbit_path)
  File "/home/users/otirmizi/tools/s1-reader/src/s1reader/s1_reader.py", line 537, in _burst_from_zip
    bursts = burst_from_xml(f_annotation, orbit_path, f_tiff, iw2_f_annotation, z_file.open)
  File "/home/users/otirmizi/tools/s1-reader/src/s1reader/s1_reader.py", line 380, in burst_from_xml
    first_valid_line = [x > 0 for x in first_valid_samples].index(True)
ValueError: True is not in list
```

This seems to be an issue arising from a handful of safe files, with the majority of them being 2016 and 2017 acquisitions for S1A

### What did you expect?

I expected [...]

### Reproducible steps

```bash
some of the zip_files =   S1A_IW_SLC__1SSV_20160829T230628_20160829T230655_012821_014394_A1F0.zip
                          S1A_IW_SLC__1SDV_20170101T002609_20170101T002639_014631_017CA0_1FAB.zip
                          S1A_IW_SLC__1SDV_20170101T002637_20170101T002705_014631_017CA0_5292.zip
                          S1A_IW_SLC__1SDV_20170113T002609_20170113T002639_014806_0181ED_ADE5.zip
                          S1A_IW_SLC__1SDV_20170125T002609_20170125T002638_014981_01875E_8F78.zip
                          S1A_IW_SLC__1SDV_20181222T002623_20181222T002653_025131_02C66C_FF54.zip
                          S1B_IW_SLC__1SDV_20190614T002608_20190614T002638_016685_01F668_3EF4.zip
                          S1B_IW_SLC__1SDV_20190626T002611_20190626T002641_016860_01FB95_3A63.zip

orbit_dir = './orbits'
[...]

PR #44, opened 2022-05-20
bug fix for first valid line

hfattahi:
This PR should fix https://github.com/opera-adt/s1-reader/issues/43

The first line or sample with index 0 can be valid. Even-though normally there are few invalid lines or samples at the beginning of each burst, but in rare cases, like those reported in  https://github.com/opera-adt/s1-reader/issues/43 the first sample of a couple of bursts starts at 0 index, ie. no invalid sample at the beginning exists.


vbrancat (approved):
@hfattahi thanks a lot for the fix!

yunjunz (approved):
I was able to reproduce the error, and confirm this PR fixed the issue.

LiangJYu (approved):
LGTM.

PR #49, opened 2022-06-24
downloading orbit

LiangJYu:
This PR puts the thinnest of wrappers on @scottstanie [sentineleof module](https://github.com/scottstanie/sentineleof) to download EOF modules from ESA or ASF.

The following script demonstrates usage:
```python
from s1reader.s1_orbit import download_orbit_file

safe_no_ext = 'S1A_IW_SLC__1SDV_20200511T135117_20200511T135144_032518_03C421_7768.'

# test both zip file and SAFE directory
for ext in ['zip', 'SAFE']:
    downloaded = download_orbit_file(f'{safe_no_ext}{ext}')
    print(downloaded)
```

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -7,3 +7,4 @@ dependencies:
   - gdal
   - isce3
   - shapely
+  - sentineleof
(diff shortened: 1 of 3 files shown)


LiangJYu:
Closing in lieu of a solution with fewer dependencies.

PR #50, opened 2022-06-30
Run pytest unit tests on circleci

rtburns-jpl:
This puts the circleci job at parity with the github action, so the github action is also removed in this PR.


.circleci/config.yml
```diff
@@ -18,6 +18,14 @@ jobs:
           name: "Download S1 test data"
           command: |
             python3 tests/download_data.py
```
yunjunz, line 20: This downloading of real S1 data is taking a long time, and is currently not used anywhere. How about commenting it out for now?

rtburns-jpl: Sounds good, I disabled it.

yunjunz (approved):
LGTM. Thank you @rtburns-jpl.

LiangJYu (approved):
LGTM. Nice work!

PR #52, opened 2022-07-08
Add orbit download functionality

vbrancat:
This PR adds to the reader the capability of downloading orbits (precise or restituted) given a SAFE file.
It requires `requests` as a dependency and orbits are downloaded from SciHub using guest credentials.

The code was originally included in the Geocoded CSLC burst stack processor but it is more appropriate to attribute orbit download to the s1-reader

### DIFF
--- a/src/s1reader/s1_orbit.py
+++ b/src/s1reader/s1_orbit.py
@@ -14,6 +14,37 @@ FMT = "%Y%m%dT%H%M%S"
 scihub_user = 'gnssguest'
 scihub_password = 'gnssguest'
 
+
+def download_orbit(safe_file, orbit_dir):
+    '''
+    Download orbits for S1-A/B SAFE "safe_file"
+
+    Parameters
+    ----------
+    safe_file: str
+        File path to SAFE file for which download the orbits
+    orbit_dir: str
+        File path to directory where to store downloaded orbits
+    '''
+
+    # Create output directory & check internet connection
+    os.makedirs(orbit_dir, exist_ok=True)
+    check_internet_connection()
+
+    # Parse info from SAFE file name
+    sensor_id, _, start_time, end_time, _ = parse_safe_filename(safe_file)
+
+    # Find precise orbit first
+    orbit_dict = get_orbit_dict(sensor_id, start_time,
+                                end_time, 'AUX_POEORB')
+    # If orbit dict is empty, find restituted orbits
+    if orbit_dict == None:
+        orbit_dict = get_orbit_dict(sensor_id, start_time,
+                                    end_time, 'AUX_RESORB')
+    # Download orbit file
+    download_orbit_file(orbit_dir, orbit_dict['orbit_url'])
+
+
 def check_internet_connection():
     '''
     Check connection availability
@@ -42,11 +73,18 @@ def parse_safe_filename(safe_filename):
        start_datetime: acquisition start datetime
        stop_datetime: acquisition stop datetime
        abs_orbit_num: absolute orbit number
+
+    Examples:
+    ---------
+    parse_safe_filename('S1A_IW_SLC__1SDV_20150224T114043_20150224T114111_004764_005E86_AD02.SAFE')
+    returns
+    ['A', 'IW', datetime.datetime(2015, 2, 24, 11, 40, 43),\
+    datetime.datetime(2015, 2, 24, 11, 41, 11), 4764]
     '''
 
     safe_name = os.path.basename(safe_filename)
     sensor_id = safe_name[2]
-    sensor_mode = safe_name[4:10]
+    sensor_mode = safe_name[4:6]
     start_datetime = datetime.datetime.strptime(safe_name[17:32],
                                                 FMT)
     end_datetime = datetime.datetime.strptime(safe_name[33:48],
@@ -134,36 +172,6 @@ def download_orbit_file(output_folder, orbit_url):
     orbit_filename = os.path.join(output_folder, header_params['filename'])
     # Save orbits
     open(orbit_filename, 'wb').write(response.content)
-
-
-def download_orbit(safe_file, orbit_dir):
-    '''
-    Download orbits for S1-A/B SAFE "safe_file"
-
-    Parameters
-    ----------
-    safe_file: str
-        File path to SAFE file for which download the orbits
-    orbit_dir: str
-        File path to directory where to store downloaded orbits
-    '''
-
-    # Create output directory & check internet connection
-    os.makedirs(orbit_dir, exist_ok=True)
-    check_internet_connection()
-
-    # Parse info from SAFE file name
-    sensor_id, _, start_time, end_time, _ = parse_safe_filename(safe_file)
-
-    # Find precise orbit first
-    orbit_dict = get_orbit_dict(sensor_id, start_time,
-                                end_time, 'AUX_POEORB')
-    # If orbit dict is empty, find restituted orbits
-    if orbit_dict == None:
-        orbit_dict = get_orbit_dict(sensor_id, start_time,
-                                    end_time, 'AUX_RESORB')
-    # Download orbit file
-    download_orbit_file(orbit_dir, orbit_dict['orbit_url'])
 
 
 def get_file_name_tokens(zip_path: str) -> [str, list[datetime.datetime]]:
--- a/src/s1reader/s1_orbit.py
+++ b/src/s1reader/s1_orbit.py
@@ -5,19 +5,14 @@ import os
 import requests
 import warnings
 
+from xml.etree import ElementTree
+
 # date format used in file names
 FMT = "%Y%m%dT%H%M%S"
 
-# Required for orbit download
-scihub_url = 'https://scihub.copernicus.eu/gnss/odata/v1/Products'
-# Namespaces of the XML file returned by the S1 query. Will they change it?
-w3_url = '{http://www.w3.org/2005/Atom}'
-m_url = '{http://schemas.microsoft.com/ado/2007/08/dataservices/metadata}'
-d_url = '{http://schemas.microsoft.com/ado/2007/08/dataservices}'
 # Scihub guest credential
 scihub_user = 'gnssguest'
 scihub_password = 'gnssguest'
-
 
 def check_internet_connection():
     '''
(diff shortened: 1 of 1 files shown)


src/s1reader/s1_orbit.py
```diff
@@ -1,12 +1,169 @@
...
 import warnings
 
-
 # date format used in file names
 FMT = "%Y%m%dT%H%M%S"
 
+# Required for orbit download
+scihub_url = 'https://scihub.copernicus.eu/gnss/odata/v1/Products'
+# Namespaces of the XML file returned by the S1 query. Will they change it?
+w3_url = '{http://www.w3.org/2005/Atom}'
```
LiangJYu, line 14: Is this needed? Global gets overwritten [here](https://github.com/opera-adt/s1-reader/blob/970ea7ed2d4aa56fe642d63010a7b7db09efb63c/src/s1reader/s1_orbit.py#L100).

src/s1reader/s1_orbit.py
```diff
@@ -1,12 +1,169 @@
...
+m_url = '{http://schemas.microsoft.com/ado/2007/08/dataservices/metadata}'
+d_url = '{http://schemas.microsoft.com/ado/2007/08/dataservices}'
```
LiangJYu, line 16: Since `scihub_url`, `m_url`, and `d_url` are only used in [`get_orbit_dict`](https://github.com/opera-adt/s1-reader/blob/970ea7ed2d4aa56fe642d63010a7b7db09efb63c/src/s1reader/s1_orbit.py#L64), let's move them there; Unless there's a reason for them to be global and importable elsewhere.

src/s1reader/s1_orbit.py
```diff
@@ -1,12 +1,169 @@
...
+
+def check_internet_connection():
+    '''
+    Check connection availability
+    '''
+    url = "http://google.com"
+    try:
+        requests.get(url, timeout=10)
+    except (requests.ConnectionError, requests.Timeout) as exception:
+        raise sys.exit(exception)
```
LiangJYu, line 30:
```suggestion
        raise ConnectionError(f'unable to reach {url}')
```

src/s1reader/s1_orbit.py
```diff
@@ -1,12 +1,169 @@
...
+
+
+def parse_safe_filename(safe_filename):
```
LiangJYu, line 33: There's enough overlap/redundancy between `parse_safe_filename` and [`get_file_name_tokens`](https://github.com/opera-adt/s1-reader/blob/970ea7ed2d4aa56fe642d63010a7b7db09efb63c/src/s1reader/s1_orbit.py#L167) that should be combined. May be better to put this in a catch-all src/s1reader/utils.py for easy importing/sharing elsewhere?

src/s1reader/s1_orbit.py
```diff
@@ -1,12 +1,169 @@
...
+        Python dictionary with [orbit_name, orbit_type, download_url]
+    '''
+    # Check if correct orbit_type
+    if orbit_type not in ['AUX_POEORB', 'AUX_RESORB']:
+        err_msg = f'{orbit_type} not a valid orbit type'
+        raise ValueError(err_msg)
+
+    # Add a 30 min margin to start_time and end_time
+    pad_start_time = start_time - datetime.timedelta(hours=0.5)
+    pad_end_time = end_time + datetime.timedelta(hours=0.5)
```
LiangJYu, line 88:
```suggestion
    pad_30_min = datetime.timedelta(hours=0.5)
    pad_start_time = start_time - pad_30_min
    pad_end_time = end_time + pad_30_min
```
Bit more clarity

src/s1reader/s1_orbit.py
```diff
@@ -1,12 +1,169 @@
...
+       abs_orbit_num: absolute orbit number
+    '''
+
+    safe_name = os.path.basename(safe_filename)
+    sensor_id = safe_name[2]
+    sensor_mode = safe_name[4:10]
+    start_datetime = datetime.datetime.strptime(safe_name[17:32],
+                                                '%Y%m%dT%H%M%S')
+    end_datetime = datetime.datetime.strptime(safe_name[33:48],
+                                              '%Y%m%dT%H%M%S')
```
LiangJYu, line 58:
```suggestion
    start_datetime = datetime.datetime.strptime(safe_name[17:32], FMT)
    end_datetime = datetime.datetime.strptime(safe_name[33:48], FMT)
```
`FMT` is global used by functions in this module.

src/s1reader/s1_orbit.py
vbrancat: Not sure whether we will need this functionality to be imported elsewhere. I am fine with the combination of the two functions. Can the combined function have as returned value the one returned by `parse_safe_filename`?

src/s1reader/s1_orbit.py
```diff
@@ -1,12 +1,171 @@
...
+    Returns:
+    -------
+    List of [sensor_id, mode_id, start_datetime,
+                end_datetime, abs_orbit_num]
+       sensor_id: sensor identifier (S1A or S1B)
+       mode_id: mode/beam (e.g. IW)
+       start_datetime: acquisition start datetime
+       stop_datetime: acquisition stop datetime
+       abs_orbit_num: absolute orbit number
+    '''
```
scottstanie, line 45: An example could be nice here at the end of the docstring, something like

```python
    Examples:
    ---------
    >>> parse_safe_filename('S1A_IW_SLC__1SDV_20150224T114043_20150224T114111_004764_005E86_AD02.SAFE')
    ['A', 'IW', datetime.datetime(2015, 2, 24, 11, 40, 43),\
datetime.datetime(2015, 2, 24, 11, 41, 11), 4764]
```

src/s1reader/s1_orbit.py
```diff
@@ -1,12 +1,171 @@
...
+
+    safe_name = os.path.basename(safe_filename)
+    sensor_id = safe_name[2]
+    sensor_mode = safe_name[4:10]
```
scottstanie, line 49: After I wrote out that example doctest, I saw that the `sensor_mode` actually doesn't match just "IW". right now it's returning "IW_SLC". 
Do we want both of those? Or should we change it to `safe_name[4:6]`?

src/s1reader/s1_orbit.py
```diff
@@ -1,12 +1,171 @@
...
+                                end_time, 'AUX_POEORB')
+    # If orbit dict is empty, find restituted orbits
+    if orbit_dict == None:
+        orbit_dict = get_orbit_dict(sensor_id, start_time,
+                                    end_time, 'AUX_RESORB')
+    # Download orbit file
+    download_orbit_file(orbit_dir, orbit_dict['orbit_url'])
+
+
 def get_file_name_tokens(zip_path: str) -> [str, list[datetime.datetime]]:
```
scottstanie, line 169: If you wanted to still keep this same function while having `parse_safe_filename` as the main parser, you could make it just
```python
def get_file_name_tokens(zip_path: str) -> [str, list[datetime.datetime]]:
    platform_id, _, start_time, end_time, _ = parse_safe_filename(zip_path)
    return platform_id, [start_time, end_time]
```

src/s1reader/s1_orbit.py
```diff
@@ -1,12 +1,171 @@
...
+    check_internet_connection()
+
+    # Parse info from SAFE file name
+    sensor_id, _, start_time, end_time, _ = parse_safe_filename(safe_file)
+
+    # Find precise orbit first
+    orbit_dict = get_orbit_dict(sensor_id, start_time,
+                                end_time, 'AUX_POEORB')
+    # If orbit dict is empty, find restituted orbits
+    if orbit_dict == None:
```
scottstanie, line 162: I'm getting a flake8 warning: `(E711) comparison to None should be 'if cond is None:'`, as in `if orbit_dict is None`

src/s1reader/s1_orbit.py
```diff
@@ -1,12 +1,171 @@
...
+    ----------
+    sensor_id: str
+        Sentinel satellite identifier ('A' or 'B')
+    start_time: datetime object
+        Sentinel start acquisition time
+    end_time: datetime object
+        Sentinel end acquisition time
+    orbit_type: str
+        Type of orbit to download (AUX_POEORB: precise, AUX_RESORB: restituted)
+    Returns:
```
scottstanie, line 72: Maybe 
```
Returns
-------
```
 just for consistency

src/s1reader/s1_orbit.py
```diff
@@ -1,12 +1,171 @@
...
+
+    # Find precise orbit first
+    orbit_dict = get_orbit_dict(sensor_id, start_time,
+                                end_time, 'AUX_POEORB')
+    # If orbit dict is empty, find restituted orbits
+    if orbit_dict == None:
+        orbit_dict = get_orbit_dict(sensor_id, start_time,
+                                    end_time, 'AUX_RESORB')
+    # Download orbit file
+    download_orbit_file(orbit_dir, orbit_dict['orbit_url'])
```
scottstanie, line 166: Do we want to add a check here that the file already exists so we don't redownload things unnecessarily? Could be 
```python
    filename = os.path.join(orbit_dir, orbit_dict["orbit_name"])
    if not os.path.exists(filename):
        download_orbit_file(orbit_dir, orbit_dict["orbit_url"])
```

scottstanie (approved):
Nice work Virginia! looks good and straightforward just like we want

src/s1reader/s1_orbit.py
vbrancat: Good catch. We do not need the product level :)

src/s1reader/s1_orbit.py
vbrancat: I like that :)

vbrancat:
@LiangJYu I think all your comments should have been addressed. I would be very glad if you can take another look earliest at your convenience

src/s1reader/s1_orbit.py
LiangJYu: I think resolving the overlapped functionality should be in different PR. I will open an issue. IIRC the noise and calibration reader has implemented it's own as well. Reminds me of [this scene](https://youtu.be/Et4sMJP9FmM?t=128) from the movie Contact.

LiangJYu (approved):
LGTM

PR #54, opened 2022-07-20
add `version.py` as the single-source version history

yunjunz:
Similar to https://github.com/opera-adt/COMPASS/pull/36, so that we only need to update version.py while cutting for a new release.

PR #56, opened 2022-07-22
setup: access version w/o __init__ to avoid circular imports

yunjunz:
This PR fixes the pip install error introduced in #54. The detailed changes are as below:

+ `setup.cfg`: setup the version number via `version.release_version`, instead of going through `__init__`, to avoid the circular imports that causes the pip install "ModuleNotFoundError"
+ `setup.cfg`: add `install_requires` and remove `[opptions.package_data]` to be consistent with compass
+ `__init__.py`: use absolute module import, to be consistent with the good style from pyresample module
+ `version.py`: remove the duplicated `__version__` variable.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,7 +16,7 @@ classifiers =
 package_dir =
     = src
 packages = find:
-include_package_data = true
+include_package_data = True
 python_requires = >=3.9
 install_requires =
     setuptools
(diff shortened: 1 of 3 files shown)


vbrancat (approved):
Works on my side. Thanks @yunjunz

seongsujeong (approved):
Works on docker as well. Thanks for the fix!

Issue #51, opened 2022-07-06
[Bug]: Loading bursts from S1 SAFE data processed by IPF 2.36 (and possibly earlier versions)


seongsujeong:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

I tried to load bursts from SAFE .zip file processed by IPF version 2.36. Running the code was not successful with the error message below:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/compass_user/miniconda3/envs/COMPASS/lib/python3.9/site-packages/s1reader/s1_reader.py", line 474, in load_bursts
    bursts = _burst_from_safe_dir(path, id_str, orbit_path)
  File "/home/compass_user/miniconda3/envs/COMPASS/lib/python3.9/site-packages/s1reader/s1_reader.py", line 583, in _burst_from_safe_dir
    bursts = burst_from_xml(f_annotation, orbit_path, f_tiff, iw2_f_annotation)
  File "/home/compass_user/miniconda3/envs/COMPASS/lib/python3.9/site-packages/s1reader/s1_reader.py", line 309, in burst_from_xml
    az_fm_rate_list = [parse_polynomial_element(x, poly_name) for x in az_rate_list_element]
  File "/home/compass_user/miniconda3/envs/COMPASS/lib/python3.9/site-packages/s1reader/s1_reader.py", line 309, in <listcomp>
    az_fm_rate_list = [parse_polynomial_element(x, poly_name) for x in az_rate_list_element]
  File "/home/compass_user/miniconda3/envs/COMPASS/lib/python3.9/site-packages/s1reader/s1_reader.py", line 52, in parse_polynomial_element
    coeffs = [float(x) for x in elem.find(poly_name).text.split()]

```

### What did you expect?

I expected a list of bursts as output. The reader worked well with the data processed by IPF 3.31

### Reproducible steps

```bash
The below is the code that I tried to load bursts from SAFE .zip file. The data is processed by IPF version 2.36.


path_safe='S1A_IW_SLC__1SSV_20150114T015005_20150114T015035_004161_0050BB_87F4.zip'
path_orbit='S1A_OPER_AUX_POEORB_OPOD_20150204T122622_V20150113T225944_20150115T005944.EOF'
bursts=s1reader.load_bursts(path_safe, path_orbit, 1, 'vv')


The sample data above is stored in the internal server.
```

### Environment

```bash
Tested environments:
- MacOS
- Linux (by @vbrancat)
[...]

PR #59, opened 2022-08-08
Parse azimuth fm rate polynomial

seongsujeong:
This PR is in regards to #51 when loading azimuth FM rate polynomial from LADS .xml files. It was found that some S1 SAFE data with old IPF has different format for the polynomial. The examples are as below:

S1A_IW_SLC__1SSV_20150114T015005_20150114T015035_004161_0050BB_87F4 (IPF `2.36`)

```xml
<azimuthFmRate>
    <azimuthTime>2015-01-14T01:50:05.584900</azimuthTime>
    <t0>5.339831264815916e-03</t0>
    <c0>-2.327628769423934e+03</c0>
    <c1>4.493262758608217e+05</c1>
    <c2>-7.852728113753408e+07</c2>
</azimuthFmRate>
```

S1B_IW_SLC__1SDV_20170807T010035_20170807T010102_006827_00C043_B7E0 (IPF `2.82`)

```xml
<azimuthFmRate>
    <azimuthTime>2017-08-07T01:00:35.780368</azimuthTime>
     <t0>5.334643585041154e-03</t0>
    <azimuthFmRatePolynomial count="3">-2.329768045614056e+03 4.502942424994464e+05 -7.864570826763810e+07</azimuthFmRatePolynomial>
</azimuthFmRate>
```

This fix detects if the field `azimuthFmRatePolynomial` exists in the input ET, and try different parsing strategy depending on the existence.


gshiroma (approved):
LGTM. Thank you for the update, @seongsujeong . I just have some minor comments:  please avoid these types of qualitative statements, especially in CAPS LOCK: `VERY old`. Also, please follow PEP8 rules for inline comments (two spaces before and a single space after `#`): https://peps.python.org/pep-0008/#inline-comments .

src/s1reader/s1_reader.py
```diff
@@ -56,7 +56,12 @@ def parse_polynomial_element(elem, poly_name):
 
     half_c = 0.5 * isce3.core.speed_of_light
     r0 = half_c * float(elem.find('t0').text)
-    coeffs = [float(x) for x in elem.find(poly_name).text.split()]
+
+    if elem.find(poly_name) is None:  # some data with old IPF version (e.g. 2.36)
```
vbrancat, line 60: We are usually leaving comments at the top of each line of code, to make sure to be in the limits of the pep8 convention.

vbrancat, line 60: I would probably say "with IPF < 2.36"

vbrancat (approved):
LGTM, just some nit-picking :)

src/s1reader/s1_reader.py
seongsujeong: @vbrancat Thanks for taking a look at the PR.
What I currently know about the relationship between IPF version and the format difference (for AZ FM rate polynomial) is that the change has taken place somewhere >2.36 and <2.82 (by taking a look at the S1 data I've processed so far). I took a look at the three different versions of S1 Product specification, but was not able to find a clear answer about the exact IPF version from which the format has changed.
I've put more context and description to the comment.

seongsujeong: Thanks for our suggestion. I've checked the style guide (https://peps.python.org/pep-0008/#inline-comments), and concluded that the inline comments would be helpful to describe which line is for the old format (and which one is for the new format). I've made sure that it passed the codacy.

PR #62, opened 2022-08-10
Thermal noise Correction and EAP phase compensation

seongsujeong:
This PR is to add thermal noise correction and EAP phase compensation features into OPERA workflow.

- Few more enhancements implemented for exporting LUT for thermal noise correction to deal with the format difference of NADS w.r.t. IPF version
- Prototype implementation for EAP correction


src/s1reader/s1_reader.py
```diff
@@ -349,29 +374,46 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
+    # load the Noise annotation
     noise_annotation_path = annotation_path.replace('annotation/', 'annotation/calibration/noise-')
     with open_method(noise_annotation_path, 'r') as f_nads:
         tree_nads = ET.parse(f_nads)
         noise_annotation = s1_annotation.NoiseAnnotation.from_et(tree_nads, ipf_version)
 
+    # load AUX_CAL annotation
+    flag_apply_eap = is_eap_correction_necessary(ipf_version)
+    if flag_apply_eap.phase_correction:
+        path_aux_cal = get_path_aux_cal(f'{os.path.dirname(s1_annotation.__file__)}/data/aux_cal',
```
LiangJYu, line 404: Will the `aux_cal` data be added in another PR?

src/s1reader/s1_reader.py
```diff
@@ -349,29 +374,46 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
+                                        platform_id,
+                                        product_annotation.inst_config_id)
+        
+        with zipfile.ZipFile(path_aux_cal,'r') as zipfile_aux_cal:
+            str_safe_aux_cal = os.path.basename(path_aux_cal).replace('.zip','')
+            with zipfile_aux_cal.open(f'{str_safe_aux_cal}/data/s1a-aux-cal.xml','r') as f_aux_cal:
+                tree_aux_cal = ET.parse(f_aux_cal)
+        aux_cal_subswath = s1_annotation.AuxCal.from_et(tree_aux_cal, pol, subswath_id)
```
LiangJYu, line 412: What do you think about creating a `load_from_zip_file` classmethod in `AuxCal` that contains all the zip file handling? I think something like that would help declutter an already crowded `s1_reader.burst_from_xml`.

Not relevant to this PR: the same logic could be applied to all the other annotations readers above.

src/s1reader/s1_reader.py
```diff
@@ -349,29 +374,46 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
         ipf_version = get_ipf_version(tree_manfest)
 
-    #Load the Product annotation - for EAP calibration
-    #with open_method(annotation_path, 'r') as f_lads:
-    #    tree_lads = ET.parse(f_lads)
-    #    product_annotation = s1_annotation.ProductAnnotation.from_et(tree_lads)
+    # Load the Product annotation - for EAP calibration
+    with open_method(annotation_path, 'r') as f_lads:
+        tree_lads = ET.parse(f_lads)
+        product_annotation = s1_annotation.ProductAnnotation.from_et(tree_lads)
```
LiangJYu, line 386:
```suggestion
        product_annotation = ProductAnnotation.from_et(tree_lads)
```
What do you think about changing imports up top so `*Annotation` doesn't need to be accessed by `s1_annotation`. The changes up top would be something like:
```
import s1_annotation # to access __file__
from s1_annotation import NoiseAnnotation, ProductionAnnotation
```

src/s1reader/s1_annotation.py
```diff
@@ -457,63 +484,146 @@ def from_calibration_annotation(cls, calibration_annotation: CalibrationAnnotati
...
-        cls.tau_sub = product_annotation.antenna_pattern_slant_range_time[id_closest]
-        cls.theta_am = product_annotation.antenna_pattern_elevation_angle
+        Ns = product_annotation.number_of_samples
+        fs = product_annotation.range_sampling_rate
+        eta_start = azimuth_time
+        tau_0 = product_annotation.slant_range_time
+        tau_sub = product_annotation.antenna_pattern_slant_range_time[id_closest]
+        #theta_sub = product_annotation.antenna_pattern_elevation_pattern[id_closest]
+        theta_sub = product_annotation.antenna_pattern_elevation_angle[id_closest]
+        #self.theta_am = product_annotation.antenna_pattern_elevation_angle
```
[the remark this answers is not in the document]
seongsujeong: Removing commented out code

src/s1reader/s1_annotation.py
```diff
@@ -457,63 +484,146 @@ def from_calibration_annotation(cls, calibration_annotation: CalibrationAnnotati
...
+    tau_0: float  # imageInformation/slantRangeTime
+    tau_sub: np.ndarray  # antennaPattern/slantRangeTime
+    theta_sub: np.ndarray  # antennaPattern/elevationAngle
     azimuth_time: datetime.datetime
+    ascending_node_time: datetime.datetime
 
     #from AUX_CAL
-    G_eap: np.ndarray #elevationAntennaPattern
-    delta_theta:float #elavationAngleIncrement
+    G_eap: np.ndarray  # elevationAntennaPattern
```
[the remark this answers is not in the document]
seongsujeong: Thanks Liang. I've also revised the other variable names (i.e. fs, Ns) into snake case.

src/s1reader/s1_annotation.py
```diff
@@ -457,63 +484,146 @@ def from_calibration_annotation(cls, calibration_annotation: CalibrationAnnotati
...
+        h_0 = 707714.8  #;m
+
+        ####Perturbation amplitudes
+        h = np.array([8351.5, 8947.0, 23.32, 11.74]) #;m
+
+        ####Perturbation phases
+        phi = np.array([3.1495, -1.5655 , -3.1297, 4.7222]) #;radians
+
+        ###Orbital time period in seconds
+        Torb = (12*24*60*60) / 175.
```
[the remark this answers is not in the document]
seongsujeong: `175` is the total number of orbits in one cycle (12 days), so `Torb` means the period to complete one orbit.

src/s1reader/s1_burst_slc.py
```diff
@@ -621,3 +622,62 @@ def swath_name(self):
...
+
+        arr_lut_total = np.matmul(az_lut_interp, rg_lut_interp)
+
+        return arr_lut_total
+
+    @property
+    def eap_compensation_lut(self):
+        '''Returns LUT for EAP compensation.
+        Based on ESA docuemnt :
+        "Impact of the Elevation Antenna Pattern Phase Compensation on the Interferometric Phase Preservation"
```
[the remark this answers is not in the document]
seongsujeong: The URL for the document has added

src/s1reader/s1_reader.py
seongsujeong: The discussion is ongoing (#61) for how to access `aux_cal`, but for now I'm assuming we are saving the data in the directory above.

src/s1reader/s1_reader.py
seongsujeong: I think that is a good idea. I've implemented `s1_annotation.AucCal.load_from_zip_file()`

>Not relevant to this PR: the same logic could be applied to all the other annotations readers above.
I'll keep that in mind by opening an issue tracker #67

src/s1reader/s1_reader.py
seongsujeong: I think that's helpful in simplifying the code, and improve the readability.
I've also imported to the classes below:
```python
ProductAnnotation
NoiseAnnotation
CalibrationAnnotation
AuxCal
BurstCalibration
BurstEAP
BurstNoise
```

seongsujeong:
Thank you @LiangJYu for reviewing this PR. I've accepted / applied your suggestions and comments as above.

src/s1reader/s1_burst_slc.py
```diff
@@ -622,3 +623,76 @@ def swath_name(self):
         '''Swath name in iw1, iw2, iw3.'''
         return self.burst_id.split('_')[1]
 
+    @property
+    def thermal_noise_lut(self):
```
LiangJYu, line 627: I think `thermal_noise_lut` would be better as class members instead as a `@property`. Identical computations would be performed each time this property is called without any change to the parameters, so I think this would be more efficient as a class member that's computed one-time. I also think the computations here would also be better suited as a member function of `BurstNoise` that takes shape as a parameter given `burst_noise` members are extensively utilized.

The changes would be something like:

s1_burst_slc.py
```
...
@dataclass(frozen=True)
class Sentinel1BurstSlc:
    ...
    thermal_noise_lut: np.array
```

s1_reader.py
```
...
def burst_from_xml()
    ...
    thermal_noise_lut = burst_noise.compute_thermal_noise_lut(shape)
    ...
    bursts[i] = Sentinel1BurstSlc(..., thermal_noise_lut, ...)
```

s1_annotation.py
```
...
@dataclass
class BurstNoise:
    ...
    def computer_thermal_noise_lut(self, shape):
        <computations highly similar to below>
...
```

Let me know what you think.

src/s1reader/s1_burst_slc.py
```diff
@@ -622,3 +623,76 @@ def swath_name(self):
...
+                                                             k=1)
+            vec_az = np.arange(self.burst_noise.line_from, self.burst_noise.line_to + 1)
+            az_lut_interpolated = az_lut_interp_obj(vec_az).reshape((nrows, 1))
+
+        arr_lut_total = np.matmul(az_lut_interpolated, rg_lut_interpolated)
+
+        return arr_lut_total
+
+    @property
+    def eap_compensation_lut(self):
```
LiangJYu, line 664: Comment regarding `thermal_noise_lut` also applicable to `eap_compensation_lut`.

src/s1reader/s1_annotation.py
```diff
@@ -267,23 +291,26 @@ class AuxCal(AnnotationBase):
...
     @classmethod
-    def from_et(cls,et_in: ET, pol: str, str_swath: str):
+    def load_from_zip_file(cls, path_aux_cal_zip: str, pol: str, str_swath: str):
         '''A class method that Extracts list of information AUX_CAL from the input ET.
 
         Parameters
         ---------
-        et_in : ET
-            ET from AUX_CAL
+        et_in : path_aux_cal_zip
```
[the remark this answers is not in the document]
seongsujeong: Thanks for catching this. I'll accept your suggested changes

src/s1reader/s1_annotation.py
```diff
@@ -457,63 +501,165 @@ def from_calibration_annotation(cls, calibration_annotation: CalibrationAnnotati
...
+        delta_anx: float
+            elapsed time from ascending node crossing time
+
+        Returns
+        -------
+        _: float
+            Estimated nominal roll (degrees)
+        '''
+        # Estimate altitude based on time elapsed since ANX
+        altitude = self._anx2height(delta_anx)
```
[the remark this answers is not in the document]
seongsujeong: @LiangJYu Good point. @hfattahi Is there any unit convention for altitude (or satellites' coordinates) for OPERA or ISCE3? I will make it consistent throughout this PR.

src/s1reader/s1_burst_slc.py
seongsujeong: @LiangJYu Very good point. I agree with your idea that it makes more sense to calculate the LUT only once.

My concern is that we turn on or off the thermal noise correction by user input in `runconfig`. I expect we will apply the correction most of the time, but we will need to turn it off when requested (for testing, comparison, etc.).
In that case, calculating the LUT is not necessary.

We can just set `thermal_noise_lut` to `None` when not applying the correction, and use it when initializing `Sentinel1BurstSlc` in `burst_from_xml()`. However, `burst_from_xml()` needs information whether user has opted in/out the correction. Maybe `s1-reader` need to access `runconfig`, but not sure if that is a good idea in terms of the independency of `s1-reader` to the workflow implementations.

I know you are not a fan of load-on-access pattern like below, but I think it could be one of the ways to optimize. Looking forward to hear about your opinion.

For your information: in our current RTC workflow, we use `thermal_noise_lut` only once when processing a burst.

s1_burst_slc.py
```python
class Sentinel1BurstSlc:
...
    _thermal_noise_lut : np.array = None

    @property
    def thermal_noise_lut(self):
        if _thermal_noise_lut is None:
            _thermal_noise_lut = <Compute the LUT>

        return _thermal_noise_lut

```


Regarding the location of the LUT computation:
> I also think the computations here would also be better suited as a member function of `BurstNoise` that takes shape as a parameter given `burst_noise` members are extensively utilized.

In user perspective, I prefer that we do not need to care about the LUT's shape by making it same as the corresponding burst's. But I agree that the LUT computation makes more sense as the method of `BurstNoise`
So I think I can move the computation part into `BurstNoise`, keep the method 
`thermal_noise_lut` inside `Sentinel1BurstSlc`, and let that method call `BurstNoise.computer_thermal_noise_lut()` by providing it with the burst's shape.

src/s1reader/s1_annotation.py
```diff
@@ -5,15 +5,16 @@
...
 
 import numpy as np
 
 from packaging import version
-from scipy.interpolate import InterpolatedUnivariateSpline
 
-#some thresholds
-version_threshold_azimuth_noise_vector=version.parse('2.90')
+#A IPF version from which has azimuth noise vector
+version_threshold_azimuth_noise_vector = version.parse('2.90')
```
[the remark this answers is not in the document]
seongsujeong: In fact, it's the version that NADS started to include azimuth noise vector. I can see that the variable's name was not clear enough to make you interpret that other way around.

My suggestions are as below:

```python
#IPF version from which the NADS has azimuth noise vector annotation
ipf_version_az_noise_vector_available_from = version.parse('2.90')
```

seongsujeong:
Realized that the changes on the code has made the CircleCI test to fail. Now working on fixing that.

seongsujeong:
> Realized that the changes on the code has made the CircleCI test to fail. Now working on fixing that.

Done.

src/s1reader/s1_annotation.py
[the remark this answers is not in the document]
LiangJYu:
```python
#IPF version from which the NADS has azimuth noise vector annotation
min_ipf_version_az_noise_vector = version.parse('2.90')
```
I think this could work too.

src/s1reader/s1_burst_slc.py
LiangJYu: Given the very significant memory usage mentioned [here](https://github.com/seongsujeong/s1-reader/pull/6#pullrequestreview-1093763303), I withdraw my suggestion of making `thermal_noise_lut` a class attribute. 

Having the `Sentinel1BurstSlc.thermal_noise_lut` property call `BurstNoise.compute_thermal_noise_lut(burst.shape)` sounds good to me.

seongsujeong: @LiangJYu Thanks for the suggestion. I've made modification to the structure.

src/s1reader/s1_burst_slc.py
seongsujeong: Same modification as `BurstNoise` has been applied for EAP compansation LUT.

src/s1reader/s1_annotation.py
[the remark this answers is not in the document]
seongsujeong: Changed the length unit as [m] for now. Will discuss this topic with the CSLC team.

src/s1reader/s1_annotation.py
```diff
@@ -457,63 +501,165 @@ def from_calibration_annotation(cls, calibration_annotation: CalibrationAnnotati
...
+        href = 711.700
+
+        # Reference boresight at reference altitude (degrees)
+        boresight_ref = 29.450
+
+        # Partial derivative of roll vs altitude (degrees/km)
+        alpha_roll = 0.0566
+
+        # Estimate nominal roll i.e. theta off nadir (degrees)
+        nominal_roll = boresight_ref - alpha_roll * (altitude/1000.0 - href)
```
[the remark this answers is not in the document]
seongsujeong: Changed the length unit as [m] for now. Will discuss this topic with the CSLC team.

src/s1reader/s1_annotation.py
```diff
@@ -138,9 +146,21 @@ class CalibrationAnnotation(AnnotationBase):
...
+        '''Extracts the list of calibration informaton from etree from CADS
+        Parameters:
+        -----------
+        et_in: ET
+            ElementTree From CADS .xml file
+
+        Returns:
+        --------
+        cls: CalibrationAnnotation
+            instance from CalibrationAnnotation initialized by the input parameter
```
gshiroma, line 159:
```suggestion
            Instance of CalibrationAnnotation initialized by the input parameter
```

src/s1reader/s1_annotation.py
```diff
@@ -152,17 +172,13 @@ def from_et(cls, et_in=None):
...
-
 @dataclass
 class NoiseAnnotation(AnnotationBase):
-    '''Reader for Noise Annotation Data Set (NADS) for IW SLC'''
-    # NOTE Schema of the NADS is slightly different before/after IPF version 2.90. Needs to be adaptive in accordance with the version.
-    # The issue above was fixed in further implement of thermal noise correction.
-    # A separete PR regarding this will be submitted upon the acceptance of this code.
-    # in ISCE2 code: if float(self.IPFversion) < 2.90:
+    '''Reader for Noise Annotation Data Set (NADS) for IW SLC
     # REF: .../isce2/components/isceobj/Sensor/GRD/Sentinel1.py
```
gshiroma, line 178:
```suggestion
```

src/s1reader/s1_annotation.py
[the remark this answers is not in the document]
seongsujeong: As discussed in last week, I will use the same units in the [original documentation](https://sentinel.esa.int/documents/247904/1653440/Sentinel-1-IPF_EAP_Phase_correction).

src/s1reader/s1_annotation.py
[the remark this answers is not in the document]
seongsujeong: Same as above. I will use the same units in the [original documentation](https://sentinel.esa.int/documents/247904/1653440/Sentinel-1-IPF_EAP_Phase_correction).

src/s1reader/s1_annotation.py
```diff
@@ -308,25 +344,33 @@ def from_et(cls,et_in: ET, pol: str, str_swath: str):
...
+                cls.azimuth_angle_increment = \
+                    float(calibration_params.find('azimuthAntennaPattern/azimuthAngleIncrement').text)
+                cls.azimuth_antenna_pattern = \
+                    np.array([float(token_val) for token_val in calibration_params.find('azimuthAntennaPattern/values').text.split()])
 
-        return cls
+                cls.azimuth_antenna_element_pattern_increment = \
+                    float(calibration_params.find('azimuthAntennaElementPattern/azimuthAngleIncrement').text)
+                cls.azimuth_antenna_element_pattern = \
+                    np.array([float(token_val) for token_val in calibration_params.find('azimuthAntennaElementPattern/values').text.split()])
```
gshiroma, line 361: Line with more than 80 chars.

src/s1reader/s1_reader.py
```diff
@@ -279,6 +285,30 @@ def get_ipf_version(tree: ET):
...
+
+    '''
+    list_aux_cal = glob.glob(f'{directory_aux_cal}/{str_platform}_AUX_CAL_V*.SAFE.zip')
+    list_aux_cal.sort()
+    path_aux_cal = list_aux_cal[instrument_cfg_id-1]
+
+    return path_aux_cal
+
 def is_eap_correction_necessary(ipf_version: version.Version) -> SimpleNamespace :
     '''Examines if what level of EAP correction is necessary, based on the IPF version.
```
gshiroma, line 313: What is EAP? Please define the acronym here.

seongsujeong:
Thank you @LiangJYu for all of your comments and suggestions. I think I've addressed all of your comments, and the PR is ready for another look if you have time. I appreciate your help :)

src/s1reader/s1_reader.py
```diff
@@ -349,28 +379,47 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
+    flag_apply_eap = is_eap_correction_necessary(ipf_version)
+    if flag_apply_eap.phase_correction:
+        path_aux_cals = f'{os.path.dirname(s1_annotation.__file__)}/data/aux_cal'
+        path_aux_cal_zip = get_path_aux_cal(path_aux_cals,
+                                            platform_id,
+                                            product_annotation.inst_config_id)
+
+        aux_cal_subswath = AuxCal.load_from_zip_file(path_aux_cal_zip,
+                                                     pol,
+                                                     subswath_id)
```
gshiroma, line 419: Does it work if EAP is not in a zip file?

src/s1reader/s1_annotation.py
```diff
@@ -259,6 +282,8 @@ def from_et(cls, et_in: ET):
         cls.number_of_samples = cls._parse_scalar('imageAnnotation/imageInformation/numberOfSamples', 'scalar_int')
         cls.range_sampling_rate = cls._parse_scalar('generalAnnotation/productInformation/rangeSamplingRate', 'scalar_float')
 
+        cls.slant_range_time =  cls._parse_scalar('imageAnnotation/imageInformation/slantRangeTime','scalar_float')
```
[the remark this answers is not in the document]
seongsujeong: Thanks for the comment. I've went through the code to make the line length <=80 characters as much as I can.

src/s1reader/s1_reader.py
seongsujeong: It's defined in the docstring in `s1_annotation.py`, but also added the definition in `s1_reader.py` as well.

src/s1reader/s1_reader.py
seongsujeong: No it will not work, and it is our intention to load AUX_CAL data from .zip file, rather than from .xml file because of the reasons described [here](https://github.com/opera-adt/s1-reader/discussions/61#discussioncomment-3647149).

seongsujeong:
@LiangJYu @gshiroma Thanks for your comments and suggestions to improve this PR. Please let me know if you have any other concerns in the code.

LiangJYu (commented):
Aside from my one comment, looks good.

src/s1reader/s1_annotation.py
```diff
@@ -277,13 +413,47 @@ class AuxCal(AnnotationBase):
     noise_calibration_factor: float
 
     @classmethod
-    def from_et(cls,et_in: ET, pol: str, str_swath: str):
-        '''A class method that Extracts list of information AUX_CAL from the input ET.
+    def get_aux_cal_instrument_config_id(cls, path_aux_cal_zip: str):
```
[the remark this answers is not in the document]
seongsujeong: @LiangJYu Thanks for testing it out. I've tested it before I commit, and just tested it again, but looks like it's working well on your side.
Would you be able to see if the codes below works on your end?

```python
import s1reader
s1reader.s1_annotation.AuxCal.get_aux_cal_instrument_config_id('S1A_AUX_CAL_V20140908T000000_G20190626T100201.SAFE.zip')
```

One thing I suspect is that S1 IW SLC data was provided. This function is for AUX_CAL file. One of the example files can be found [here](https://sar-mpc.eu/download/3c8b7c8d-d3de-4381-a19d-7611fb8734b9/):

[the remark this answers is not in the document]
LiangJYu: @seongsujeong thank you for pointing me to the correct file. Everything works now.

LiangJYu (approved):
LGTM

src/s1reader/s1_annotation.py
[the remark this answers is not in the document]
seongsujeong: @LiangJYu I think it gives a good point about guiding the user. I'll add few lines to raise error when unexpected file path is provided.

src/s1reader/s1_reader.py
```diff
@@ -349,28 +394,47 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
     # burst sensing starts in prototyping test data
     burst_interval = 2.758277
```
gshiroma, line 395: I know this is not part of this PR, but I was wondering where did the magic number above came from? Should we read it from the metadata?

src/s1reader/s1_reader.py
```diff
@@ -349,28 +394,47 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
+    # Load the Product annotation - for EAP calibration
+    with open_method(annotation_path, 'r') as f_lads:
+        tree_lads = ET.parse(f_lads)
+        product_annotation = ProductAnnotation.from_et(tree_lads)
 
-    #load the Calibraton annotation
-    calibration_annotation_path = annotation_path.replace('annotation/', 'annotation/calibration/calibration-')
+    # load the Calibraton annotation
+    calibration_annotation_path =\
+        annotation_path.replace('annotation/', 'annotation/calibration/calibration-')
```
gshiroma, line 410: You should avoid using forward slashes to separate directories. This would not work in a windows machine. Ideally, you'd do the operations from `os.path`, for example, instead of `'annotation/calibration/calibration-'`, you'd have something like `os.path.join('annotation', ',calibration', 'calibration-').

src/s1reader/s1_reader.py
```diff
@@ -349,28 +394,47 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
     with open_method(calibration_annotation_path, 'r') as f_cads:
         tree_cads = ET.parse(f_cads)
-        calibration_annotation = s1_annotation.CalibrationAnnotation.from_et(tree_cads)
+        calibration_annotation =\
+            CalibrationAnnotation.from_et(tree_cads,
+                                          calibration_annotation_path)
 
-    #load the Noise annotation
+    # load the Noise annotation
     noise_annotation_path = annotation_path.replace('annotation/', 'annotation/calibration/noise-')
```
gshiroma, line 418: Not part of this PR, but we should not use forward slashes as separators

src/s1reader/s1_reader.py
```diff
@@ -349,28 +394,47 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
     with open_method(noise_annotation_path, 'r') as f_nads:
         tree_nads = ET.parse(f_nads)
-        noise_annotation = s1_annotation.NoiseAnnotation.from_et(tree_nads, ipf_version)
+        noise_annotation = NoiseAnnotation.from_et(tree_nads, ipf_version,
+                                                   noise_annotation_path)
+
+    # load AUX_CAL annotation
+    flag_apply_eap = is_eap_correction_necessary(ipf_version)
+    if flag_apply_eap.phase_correction:
+        path_aux_cals = f'{os.path.dirname(s1_annotation.__file__)}/data/aux_cal'
```
gshiroma, line 427: After installation using conda, will `data/aux_cal` be installed as a sub-directory of `os.path.dirname(s1_annotation.__file__)`? Notice that conda may change the tree structure of your repo, separating executables, modules, and data.
Also, please use `os.path.join()` for directory/file separator rather than forward slashes.

src/s1reader/s1_reader.py
```diff
@@ -349,28 +394,47 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
+    # load the Noise annotation
     noise_annotation_path = annotation_path.replace('annotation/', 'annotation/calibration/noise-')
     with open_method(noise_annotation_path, 'r') as f_nads:
         tree_nads = ET.parse(f_nads)
-        noise_annotation = s1_annotation.NoiseAnnotation.from_et(tree_nads, ipf_version)
+        noise_annotation = NoiseAnnotation.from_et(tree_nads, ipf_version,
+                                                   noise_annotation_path)
+
+    # load AUX_CAL annotation
+    flag_apply_eap = is_eap_correction_necessary(ipf_version)
```
gshiroma, line 425: We should have a way to make this flag false if we are only interested in backscattering (e.g., RTC-S1).

gshiroma (commented):
Thank you so much, @seongsujeong, for working on this PR. This is really important for RTC and CSLC. As we discussed offline, I think we need  aux_cal in this PR. Otherwise, we'll have problems processing old data. In addition, I also added a few comments below.

src/s1reader/s1_reader.py
seongsujeong: I did not write this line, but I would guess it can be calculated from `azimuthTime` from the consecutive two bursts, so I think it can be readable from the product annotation .xml file.

src/s1reader/s1_reader.py
LiangJYu: In the context of OPERA, since workflows dependent on `s1-reader` all use `isce3`, which does not support Windows, I don't think it a good use of time and effort to accommodate Windows natively in `s1-reader`.

If someone really wants to run this on Windows, they can install WSL2 and install Conda and `s1-reader` there.

LiangJYu: @gshiroma if you think native Windows support is vital, perhaps open an issue or start a discussion in this repo? This way we don't derail discussion in this PR.

src/s1reader/s1_reader.py
LiangJYu: This number was computed in roughly the same way @seongsujeong described. The difference being the process described above was applied to many consecutive bursts and the average was taken.

src/s1reader/s1_reader.py
gshiroma: Hi @LiangJYu , yes, we are not supporting Windows. I just think it's a good practice to avoid using `/` as a directory separator. That way it would be easier if someday we decide to make certain pieces of code/modules working in other machines. Or it would be easier for other developers to do so, even if it's not our goal.

src/s1reader/s1_reader.py
gshiroma: Right, but this can change with time or different instruments (e.g., Sentinel-1C/D)?

LiangJYu: @gshiroma S1C is a problem for 2023 when we have data?

Also the explanation of the number is provided in the comments immediately [above](https://github.com/opera-adt/s1-reader/blob/71e2c9e589f9e010b4952a574fcda83806561339/src/s1reader/s1_reader.py#L345-L349).

src/s1reader/s1_reader.py
seongsujeong: @gshiroma `calibration_annotation_path` can be path inside .zip file when S1 data is provided as a .zip file. The path will be in the file system when the directory to (unzipped) SAFE directory is provided.

I've found the article below, and wanted to check how the path looks like with Windows Python, but I do not have one handy right now.

https://stackoverflow.com/questions/8176953/python-zipfile-path-separators

If that is true, we should differentiate the separator (i.e. forward- or back slash) depending on the SAFE file form (.zip or directory), as well as the platform (UNIX-like, or Windows).

src/s1reader/s1_reader.py
seongsujeong: Same as the comment in `calibration_annotation_path`

src/s1reader/s1_reader.py
gshiroma: Thank you for all the info, @LiangJYu . We just need to make sure that this code is robust globally and to potential changes to the instrument (not sure if that's possible). Hopefully it will not cause wrong burst IDs for specific burst locations due to cumulative time differences. And yes, it's something to keep in mind for S1C and S1D too. When S1C data is available, the project will probably ask us to test data from those sensors. These type of hidden details are dangerous when we don't remember about them anymore. If the S1C process runs, maybe we'll forget to check whether this `burst_interval` is the same for S1C.

That's why I asked if there was no change of reading it from the metadata? I guess you and @hfattahi are aware of that when this was written?

src/s1reader/s1_reader.py
gshiroma: From the reference you pointed out, I believe that it expects `os.sep` (which is the system separator). So, you wouldn't need to differentiate them. You can just use `os.sep` and `os.path.join`.

seongsujeong: Managed to get a Windows machine and do some test. It turned out that we use forward slash when referring to the file in a .zip file. in Windows. Therefore `os.path.join` is not going to work on Windows machine when attempting to load bursts from S1 .zip file.

I totally agree that `os.path.join` is better practice in general. But our s1-reader has an option to load directly from .zip file. in that case, we need to use `/` regardless of the platform as in the table below:

| | Unix-like |  Windows|
|----|--------|-----------|
| S1 data as .zip| / | / |
|S1 data as directory| / | \ |

This implies that it takes more effort than replacing the line with `os.path.join` to make this line compatible with other platform (i.e. Windows). Of course using the join function would work okay with our current target platform (linux), but not sure if we this is absolutely necessary.

src/s1reader/s1_reader.py
seongsujeong: `path_aux_cals` is indeed a path in the file system. Therefore I agree that it makes more sense to use `os.path.join()` here. I've modified & tested this line.

src/s1reader/s1_reader.py
seongsujeong: `flag_apply_eap` was added into the function `load_bursts()`. See the function header below:

```python
def load_bursts(path: str, orbit_path: str, swath_num: int, pol: str='vv',
                burst_ids: list[str]=None,
                flag_apply_eap: bool=True):
```

src/s1reader/s1_reader.py
```diff
@@ -279,8 +285,89 @@ def get_ipf_version(tree: ET):
...
+                id_match = i
+                dt_validation_prev = dt_validation
+                dt_generation_prev = dt_generation
+                continue
+
+            if dt_validation_prev == dt_validation:
+                # Same validity time; Choose the one with latest generation time
+                if dt_generation_prev > dt_generation:
+                    id_match = i
+                    dt_generation_prev = dt_generation
```
gshiroma, line 358: Thank you @seongsujeong for the updates. Sorry, I don't understand what the rule is here for selecting the AUX_CAL file. I thought that maybe we wanted to (1) compare when the dataset was acquired and compare with the "validation date" first, and (2) if we don't find a match, we'd take one of the AUX_CAL files that are available for that configuration ID. Maybe you are doing the latter here (2) by getting the AUX_CAL with latest "validation date", but I don't see the former (1).

seongsujeong: I've took some time to search through the related document and was able to found the selection criteria suggested by ESA, which is:

>The criteria for selecting an appropriate auxiliary product to use when processing a particular job order are as follows:
>1. Select the auxiliary product(s) with a validity start date/time closest to, but not later than, the start of the job order;
>2. If there is more than one product which meets the first criteria (e.g. two auxiliary files have the same validity date/time), then use the auxiliary product with the latest generation time.

The revised function is the implementation of the criteria above.
The link to the document is provided in the docstring.

gshiroma: That's great that you found this, @seongsujeong . It looks that your code is prioritizing the latest `dt_validation_prev`, which doesn't seem to be what they suggest? Or maybe I'm misunderstanding this. If everything is correct, could you please add some comments so it's easier to understand the code? 
Tagging @vbrancat and @LiangJYu so they can follow this discussion.

seongsujeong: `dt_validation_prev` in fact the "best `dt_validation`" that satisfies the 1st criteria above.
The algorithm here is to iterate through the found AUX_CAL files, and compare the `dt_validation` in the current round with the "most suitable" validation time that has found so far (i.e. dt_validation_prev). `dt_validation_prev` gets updated when `dt_validation` is better (and the iteration goes on).

Maybe changing `dt_validation_prev` as sth. like `dt_validation_best_so_far` would cause less confusion?

gshiroma: You are right, @seongsujeong . Thank you for all info! Would you please add the text by ESA in the comments? I think that will be very helpful to anyone reading the code. 

For RTC, I'd be a little worried about the case that you don't find a suitable AUX_CAL and the exception was raised for a correction that is not needed (at least for RTC), but I'm glad that you are adding a flag to turn it off, so this will not be a problem. Thanks! 

Also, please "early escape" the `if` statement `dt_validation<0` with a `continue`.

seongsujeong: Sounds good. I've applied early escape, and added more information on  the docstring and the comments between the lines to help in better understanding the algorithm.

src/s1reader/s1_reader.py
```diff
@@ -279,8 +285,102 @@ def get_ipf_version(tree: ET):
...
+            id_match = i
+            dt_validation_prev = dt_validation
+            dt_generation_prev = dt_generation
+            continue
+
+        if dt_validation_prev == dt_validation:
+            # Same validity time; Choose the one with latest generation time
+            if dt_generation_prev > dt_generation:
+                id_match = i
+                dt_generation_prev = dt_generation
```
gshiroma, line 371:
```suggestion
        # Same validity time; Choose the one with latest generation time
        if (dt_validation_prev == dt_validation and
                dt_generation_prev > dt_generation):
            id_match = i
            dt_generation_prev = dt_generation
```

gshiroma (approved):
Thank you, @seongsujeong , for addressing all my comments. I have one last suggestion below, but I think this PR is already good for merging.

gshiroma:
> Thank you, @seongsujeong , for addressing all my comments. I have one last suggestion below, but I think this PR is already good for merging.

Ah.. Just please make sure to resolve all conflicts before merging.

PR #63, opened 2022-08-11
version: add version 0.1.2 Tag

yunjunz:
+ version.py: add version 0.1.2 tag, to sync setup.py with the GitHub release.


LiangJYu (approved):
LGTM

yunjunz:
> Looks good to me. Maybe we should remove the tag `v0.1.2` currently in place, and re-tag this version to the new `v0.1.2`?

This is just for info used in the setup.py with pip install. I think it's fine to not re-tag. The two (version.py and GitHub release) will be synced next time when we cut for our next release.

PR #69, opened 2022-09-21
WIP noise annotation refactor

LiangJYu:
This PR builds on @seongsujeong's work from #62. In the early draft stage, f019bae below is the only relevant commit. I'm unsure if changes here should be a PR submitted to #62 or standalone PR pending #62 merging; currently a standalone for higher visibility.

This PR:
* moves noise loading into an independent module.
* only loads azimuth attributes as needed.
* creates utils.py in src/s1reader/utils for common variables/methods. (utils. py may not be the best name?)

To do:
- [ ] test code
- [ ] add unit test


vbrancat:
@lyu it is unclear to me why we need this refactor. If I am not wrong, @seongsujeong PR has not merged yet. Can we just be done with those changes in his open PR? I think I am missing or I have lost track of something.

lyu:
@vbrancat It would be appreciated if you could @ the correct username of Liang Yu in the future, this is the 3rd/4th time that you accidently @ me. Thank you.

PR #71, opened 2022-09-26
`s1_orbit.get_orbit_file_from_dir()`: add `auto_download` arg

yunjunz:
+ `s1_orbit.get_orbit_file_from_dir()`: add `auto_download` arg to automatically download the orbit file if not found. The default is False.

+ remove the unused get_orbit_file_from_list()

+ utils.plot_bursts.py: switch to get_orbit_file_from_dir()

### DIFF
--- a/src/s1reader/s1_orbit.py
+++ b/src/s1reader/s1_orbit.py
@@ -15,7 +15,7 @@ scihub_user = 'gnssguest'
 scihub_password = 'gnssguest'
 
 
-def download_orbit(safe_file, orbit_dir):
+def download_orbit(safe_file: str, orbit_dir: str):
     '''
     Download orbits for S1-A/B SAFE "safe_file"
 
@@ -25,6 +25,11 @@ def download_orbit(safe_file, orbit_dir):
         File path to SAFE file for which download the orbits
     orbit_dir: str
         File path to directory where to store downloaded orbits
+
+    Returns:
+    --------
+    orbit_file : str
+        Path to the orbit file.
     '''
 
     # Create output directory & check internet connection
@@ -42,9 +47,11 @@ def download_orbit(safe_file, orbit_dir):
         orbit_dict = get_orbit_dict(sensor_id, start_time,
                                     end_time, 'AUX_RESORB')
     # Download orbit file
-    filename = os.path.join(orbit_dir, orbit_dict["orbit_name"])
-    if not os.path.exists(filename):
-       download_orbit_file(orbit_dir, orbit_dict['orbit_url'])
+    orbit_file = os.path.join(orbit_dir, orbit_dict["orbit_name"] + '.EOF')
+    if not os.path.exists(orbit_file):
+        download_orbit_file(orbit_dir, orbit_dict['orbit_url'])
+
+    return orbit_file
 
 
 def check_internet_connection():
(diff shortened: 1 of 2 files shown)


src/s1reader/s1_orbit.py
```diff
@@ -245,50 +267,23 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list[str]) -> str:
         t_orbit_start = datetime.datetime.strptime(t_orbit_start[1:], FMT)
 
         # string '.EOF' from end of end time string
-        t_orbit_end = datetime.datetime.strptime(t_orbit_end[:-4], FMT)
+        t_orbit_stop = datetime.datetime.strptime(t_orbit_end[:-4], FMT)
```
scottstanie, line 270: Just checking- this name change was to make line 275 more logical (since the comment above there has "stop" instead of "end")?
I'm noticing now that this file alone has 26 instances of `end_*` and 11 instances of `stop_*`, and most other files have a combo of "stop" and "end". Not sure if we want to pick one and change the rest to match, but perhaps that can be another pull request after this.

scottstanie (approved):
This looks good to me, and is great timing- I was thinking about how the orbit files should download if you've only got the .zip files, so thanks for this.

src/s1reader/s1_orbit.py
yunjunz: I prefer `stop` rather than `end`, because 1) "stopTime" is used in the S1 XML file, and 2) isce2 variable names, and 3) burst member variable [here](https://github.com/opera-adt/s1-reader/blob/71e2c9e589f9e010b4952a574fcda83806561339/src/s1reader/s1_burst_slc.py#L590).

Replacing "end" with "stop" throughout `s1reader` sounds good to me. And yes, I think it's better to do that in another dedicated PR.

src/s1reader/s1_orbit.py
```diff
@@ -204,35 +218,43 @@ def get_file_name_tokens(zip_path: str) -> [str, list[datetime.datetime]]:
...
+        Path to the orbit file.
     '''
+
+    # check the existence of input file path and directory
     if not os.path.exists(zip_path):
         raise FileNotFoundError(f"{zip_path} does not exist")
 
-    if not orbit_file_list:
-        raise ValueError("no orbit paths provided")
+    if not os.path.isdir(orbit_dir):
```
scottstanie, line 247: Oh one thing I got while testing this- Should we just make the directory if we specify one that doesn't exist with a `os.makedirs(orbit_dir, exist_ok=True)`? I'm thinking of integrating this into Virginia's stack workflow, and it will make a new directory for orbits as long as you have one with the zip files https://github.com/opera-adt/COMPASS/pull/18

yunjunz: Either is fine with me, I don't have a preference for this. 

I did not do that because 1) my orbit directory always exists (since the same one is used for all the processing); 2) occasionally I make typos on the path, thus, creating-dir-if-not-exist does not suit my need.

scottstanie: ok that seems equally good to make sure they user has typed in the correct orbit directory 👍

vbrancat (approved):
LGTM

PR #73, opened 2022-10-05
add s1-etad prep module for testing S1-ETAD products

yunjunz:
This PR adds `s1_etad.py`, which is a wrapper around the `s1etad` module from ESA. Given an `s1reader.Sentinel1BurstSlc` object, this sub-module will search, read the ETAD corrections, and/or resample into the full resolution, following the `s1etad` documentation (https://s1etad.readthedocs.io/en/latest/notebooks/S1-ETAD-Product_Quick_User_Guide.html).

Below is a simple example of usage.

```python
import s1reader

orbit_dir = '~/data/aux/S1/aux_poeorb'
eta_dir = './ETA'
slc_file = './SLC/S1A_IW_SLC__1SSV_20160922T230629_20160922T230656_013171_014F06_1463.zip'

# read SLC burst
orbit_file = s1reader.get_orbit_file_from_dir(slc_file, orbit_dir)
slc_burst = s1reader.load_bursts(slc_file, orbit_file, swath_num=1)[0]

# read ETA burst
eta_burst, eta = s1reader.s1_etad.get_eta_burst_from_slc_burst(slc_burst, eta_dir, verbose=True)
```

### DIFF
--- a/src/s1reader/s1_etad.py
+++ b/src/s1reader/s1_etad.py
@@ -46,10 +46,9 @@ def get_eta_correction_from_slc_burst(slc_burst, eta_dir, corr_type='sum', inclu
     vprint = print if verbose else lambda *args, **kwargs: None
     if unit not in ['second', 'meter', 'pixel']:
         raise ValueError(f'Un-recognized input unit={unit}!')
-    elif unit == 'meter':
-        meter = True     #read s1etad product in meters
-    else:
-        meter = False    #read s1etad product in seconds [recommended]
+    meter = (unit == 'meter')
+    # When True, read s1etad product in meters
+    # When False, read s1etad product in seconds [recommended]
 
     # locate / read ETA burst
     eta_burst, eta = get_eta_burst_from_slc_burst(slc_burst, eta_dir, verbose=verbose)
--- a/src/s1reader/s1_etad.py
+++ b/src/s1reader/s1_etad.py
@@ -94,9 +94,9 @@ def get_eta_correction_from_slc_burst(slc_burst, eta_dir, corr_type='sum', inclu
         eta_rg_ax = eta_rg_start + np.arange(eta_burst.samples) * eta_burst.sampling['x']
 
         # calculate SLC grid
-        slc_az_ax = np.arange(slc_burst.length) * slc_burst.azimuth_time_interval
+        slc_az_ax = np.arange(slc_burst.length) * slc_burst.azimuth_time_interval \
                     + (slc_burst.sensing_start - slc_burst.sensing_mid).total_seconds()
-        slc_rg_ax = np.arange(slc_burst.width) / slc_burst.range_sampling_rate
+        slc_rg_ax = np.arange(slc_burst.width) / slc_burst.range_sampling_rate \
                     + slc_burst.slant_range_time
 
     if resample:
--- a/src/s1reader/s1_etad.py
+++ b/src/s1reader/s1_etad.py
@@ -39,7 +39,7 @@ def get_eta_correction_from_slc_burst(slc_burst, eta_dir, corr_type='sum', inclu
     Returns:
     ----------
     slc_rg_corr: np.ndarray in float32 in size of (lines, samples)
-        S1 ETAD correction in range   direction in the unit of meter or second
+        S1 ETAD correction in range direction in the unit of meter or second
     slc_az_corr: np.ndarray in float32 in size of (lines, samples)
         S1 ETAD correction in azimuth direction in the unit of meter or second
     """
--- a/src/s1reader/s1_etad.py
+++ b/src/s1reader/s1_etad.py
@@ -5,7 +5,6 @@ import os
 
 import numpy as np
 from matplotlib import pyplot as plt
-from scipy.constants import c
 from scipy.interpolate import RectBivariateSpline
 
 try:
--- a/src/s1reader/s1_etad.py
+++ b/src/s1reader/s1_etad.py
@@ -189,7 +189,7 @@ def get_eta_file_from_slc_burst(slc_burst, eta_dir, verbose=True):
     """Get/locate ETAD file path based on SLC burst."""
 
     # safe filename --> ETA filename pattern
-    fparts = os.path.basename(slc_burst.file_id).split('_')
+    fparts = os.path.basename(slc_burst.safe_filename).split('_')
     eta_fbase = f'{fparts[0]}_IW_ETA__*_{fparts[5]}_{fparts[6]}_{fparts[7]}_{fparts[8]}_*.SAFE'
 
     # search the ETA filename pattern
--- a/src/s1reader/s1_etad.py
+++ b/src/s1reader/s1_etad.py
@@ -1,0 +1,204 @@
+"""Wrapper of s1etad module to read S1-ETAD products."""
+import datetime
+import glob
+import os
+
+import numpy as np
+from matplotlib import pyplot as plt
+from scipy.constants import c
+from scipy.interpolate import RectBivariateSpline
+
+try:
+    import s1etad
+except ImportError:
+    raise ImportError('Can NOT import s1etad (https://gitlab.com/s1-etad/s1-etad)!')
+
+
+
+def get_eta_correction_from_slc_burst(slc_burst, eta_dir, corr_type='sum', include_tropo=True,
+                                      resample=True, plot=False, verbose=True, unit='second'):
+    """Get the ETAD correction for one S1 burst.
+
+    Parameters
+    ----------
+    slc_burst: Sentinel1BurstSlc
+        Sentinel-1 burst object from sentinel1_reader
+    eta_dir: str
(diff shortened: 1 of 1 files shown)


src/s1reader/s1_etad.py
```diff
@@ -0,0 +1,203 @@
...
+    # read ETA correction data
+    corr_type = corr_type.lower()
+    if isinstance(corr_type, list):
+        corr_types = list(corr_type)
+    elif corr_type == 'sar':
+        corr_types = ['bistatic', 'doppler', 'fmrate']
+    elif corr_type == 'atm':
+        corr_types = ['ionospheric', 'tropospheric']
+    else:
+        corr_types = [corr_type]
```
seongsujeong, line 66: - What's `corr_types` going to be when some case like `corr_type == ['sar', 'atm']`?
- Is `eta_burst.get_correction()` able to take `sum`, `sar`, or `atm` as argument for parameter `corr_type`?

seongsujeong (commented):
Looks good overall. I've left few suggestions and questions.

The title of this PR seems to tell that I should expect a notebook. Are you going to upload that in the branch?

src/s1reader/s1_etad.py
```diff
@@ -0,0 +1,203 @@
...
+    resample: bool
+        resample the low resolution ETA to the full resample SLC size
+    unit: str
+        output ETA correction unit:
+        pixel, second (meter is not recommended, as time info is more consistent and universal in SAR)
+
+    Returns:
+    ----------
+    slc_rg_corr: np.ndarray in float32 in size of (lines, samples)
+        S1 ETAD correction in range   direction in the unit of meter or second
```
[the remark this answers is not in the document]
yunjunz: I added a few whitespaces intentionally to better align with the comments below, for readability. It's fine with me to change it as you suggested, as long as we all agree with that convention. Tagging @LiangJYu to see his opinion.

src/s1reader/s1_etad.py
yunjunz: + `corr_type == ['sar', 'atm']` will stay the same, as covered in line 59-60.

> Is eta_burst.get_correction() able to take sum, sar, or atm as argument for parameter corr_type?

Yes, that's the exactly the purpose of this part.

yunjunz:
> The title of this PR seems to tell that I should expect a notebook. Are you going to upload that in the branch?

Thank you @seongsujeong. I have a notebook, but will upload it to another independent repo, as notebooks can easily increase the code repo size. I modified the PR title to avoid this confusion.

src/s1reader/s1_etad.py
```diff
@@ -0,0 +1,202 @@
...
+
+    if not resample:
+        return eta_rg_corr, eta_az_corr
+
+    return slc_rg_corr, slc_az_corr
+
+
+
+def get_eta_burst_from_slc_burst(slc_burst, eta_dir, verbose=True):
+    """Read ETA burst corresponding to the input SLC burst."""
```
vbrancat, line 156: What does ETA stand for?

src/s1reader/s1_etad.py
```diff
@@ -0,0 +1,202 @@
...
+import datetime
+import glob
+import os
+
+import numpy as np
+from matplotlib import pyplot as plt
+from scipy.interpolate import RectBivariateSpline
+
+try:
+    import s1etad
```
vbrancat, line 11: I am a bit confused here. Are we planning to add this dependency? If yes, please update the list of dependency for the s1-reader. In addition, I thought that our goal was to compute the corrections from scratch. Are we adding support for ETAD corrections files for debug purposes?

src/s1reader/s1_etad.py
```diff
@@ -0,0 +1,202 @@
...
+except ImportError:
+    raise ImportError('Can NOT import s1etad (https://gitlab.com/s1-etad/s1-etad)!')
+
+
+
+def get_eta_correction_from_slc_burst(slc_burst, eta_dir, corr_type='sum', include_tropo=True,
```
vbrancat, line 17: Does the title of function miss a D for `eta`? Through out the PR you use ETA and ETAD but I am not sure if this is done intentionally or not

src/s1reader/s1_etad.py
```diff
@@ -0,0 +1,202 @@
...
+    ----------
+    slc_rg_corr: np.ndarray in float32 in size of (lines, samples)
+        S1 ETAD correction in range   direction in the unit of meter or second
+    slc_az_corr: np.ndarray in float32 in size of (lines, samples)
+        S1 ETAD correction in azimuth direction in the unit of meter or second
+    """
+    vprint = print if verbose else lambda *args, **kwargs: None
+    if unit not in ['second', 'meter', 'pixel']:
+        raise ValueError(f'Un-recognized input unit={unit}!')
+    meter = (unit == 'meter')
```
vbrancat, line 49: I am wondering if we should add a warning when the user selects `meter` as a unit. This warning is hidden in the doc string but probably it is more visible to an unexperienced user with a print message

vbrancat (commented):
@yunjunz thanks for this PR. Will there be a Jupiter Notebook that we can use to practice the use of the ESA ETAD correction? I left some comments. I still need to test the PR. I am wondering if the `s1etad` module (installable through conda?) is going to take care of the downloading of the ETAD correction for that specific burst. Thanks

src/s1reader/s1_etad.py
yunjunz: ETA stands for extended timing annotation, it's a term used in the SAFE file/dir by ESA, similar to SLC, e.g. `S1A_IW_ETA__AXSV_20160805T230627_20160805T230654_012471_0137E2_91C6.SAFE`

src/s1reader/s1_etad.py
yunjunz: >  I thought that our goal was to compute the corrections from scratch. Are we adding support for ETAD corrections files for debug purposes?

That's my understanding currently. So I did not update the dependency of `s1reader`.

src/s1reader/s1_etad.py
yunjunz: Both ETA and ETAD are used by ESA. I am using `eta` in all func/var names, and ETAD in the comments. Changing all to `etad` sounds good to me too, let me know please.

src/s1reader/s1_etad.py
yunjunz: We recommend using second because it's more accurate than meter for SAR data record, but I don't have a number on how big of a difference it could be. Is this the reason you wanted a warning msg?

src/s1reader/s1_etad.py
vbrancat: I see, I was confused because I have always heard ETAD. Thanks for the clarification.

src/s1reader/s1_etad.py
vbrancat: I think it is a bit of a contradiction to support reading ETAD datasets but not including the dependency necessary to read them. I am not sure how to go about this as the only occasion I see this be handy is for debugging purposes (i.e. compare our in-house developed correction layers with that from ESA/ETAD). Is the `s1etad` dependency a light one? If this is the case, probably, we should just add it. What is your take on this?

src/s1reader/s1_etad.py
vbrancat: Your nomenclature is fine as it is. :)

src/s1reader/s1_etad.py
vbrancat: Yeah! I think an unexperienced user won't think about it. So, if we throw a warning, we might get its attention to that. On the other hand, this might be useless if we do not know what the discrepancy in accuracy between meters and seconds XD

vbrancat:
@yunjunz how can I test this PR? Are the ETAD product already available and downloadable? Or do you have a Notebook with a workflow? Merci

src/s1reader/s1_etad.py
yunjunz: As you have said, this module is for testing purposes only. The `s1etad` module dependency is not light. Thus, I like the current way of not adding it to our requirements.txt file. Users of this script is assume to be advanced, thus, know how to install s1etad, based on the printout msg.

yunjunz:
> @yunjunz how can I test this PR? Are the ETAD product already available and downloadable? Or do you have a Notebook with a workflow? Merci

@vbrancat with the link I sent offline, you should be able to download the S1ETAD products and test based on the note I wrote at the beginning of this PR.

yunjunz:
Here is the notebook example: https://github.com/opera-adt/notebooks/pull/1

vbrancat (approved):
Testing looked good on my side. Thanks @yunjunz :)

vbrancat:
@seongsujeong do you have time to look at this PR? It might be interesting for the work you are doing for the azimuth FM-rate mismatch

seongsujeong (approved):
LGTM. Thanks for the PR!

PR #76, opened 2022-10-24
s1_reader.py:zero pad burst_num in `.burst_id`

scottstanie:
The current format doesnt matche the assumed format of the burst database.

https://github.com/opera-adt/burst_db/blob/main/build_database.py#L164

This will always set the burst_num to be 6 digits long.

An alternative to this fix could be 
1. change the burst database, or
2. Start pulling organization-wide utility functions into a separate, shared repository so that the implementations dont diverge (like `get_point_epsg`)

### DIFF
--- a/src/s1reader/s1_reader.py
+++ b/src/s1reader/s1_reader.py
@@ -632,7 +632,7 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
                           last_valid_samples[last_line])
 
 
-        burst_id = f't{track_number:03d}_{burst_num}_{swath_name.lower()}'
+        burst_id = f't{track_number:03d}_{burst_num:06d}_{swath_name.lower()}'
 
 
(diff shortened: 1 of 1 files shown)


vbrancat:
@scottstanie The desired behavior is that the `s1-reader` and burst ID DB follow the same burst ID convention defined by ESA. Which one of the two conventions is consistent with that of ESA?

scottstanie:
In the sqlite database, the burst ID and relative orbit numbers are integers so they are not shown as zero pads. But the filenames of the KMZ files are zero padded:

```
(mapping) staniewi:kmz$ pwd
/Users/staniewi/Downloads/S1_burstid_20220530/IW/kmz
(mapping) staniewi:kmz$ ls | head
001_IW_000001_002148.kmz
002_IW_002149_004296.kmz
003_IW_004297_006443.kmz
```

scottstanie:
this is subsumed by #77

PR #79, opened 2022-11-10
Get orbit file

seongsujeong:
In context of the discussion in [here](https://github.com/opera-adt/COMPASS/issues/54), this PR will make the reader to take different formats of arguments for orbit. 
Below are the highlights of the changes:

- `get_orbit_file_from_list()` is revived. Most of the lines are originated from the original `get_orbit_file_from_dir()`
- `get_orbit_file_from_dir()` is modified to get the list of orbit files in the directory provided. Internally, the list will be passed on `get_orbit_file_from_list()` to pick the orbit file that corresponds to SAFE file.
- Automatic download is attempted only for `get_orbit_file_from_dir()`


src/s1reader/s1_orbit.py
```diff
@@ -278,12 +332,8 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
             orbit_file = ''
```
gshiroma, line 332: I know this is not from your PR. I personally don't like to use the variable `orbit_file` later in the code from a variable that is also used for the `for` loop, but I guess it's OK. In any case, this `else` is not necessary. The assignment `orbit_file = ''` can be done right after the break.

src/s1reader/s1_orbit.py
```diff
@@ -278,12 +332,8 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
...
 
     if not orbit_file:
-        if auto_download:
-            orbit_file = download_orbit(zip_path, orbit_dir)
-
-        else:
-            msg = f'No orbit file found for {os.path.basename(zip_path)}!'
-            msg += f'\nOrbit directory: {orbit_dir}'
-            warnings.warn(msg)
+        msg = 'No orbit file found from the file list provided.!'
```
gshiroma, line 335:
```suggestion
        msg = 'No orbit file was found in the file list provided.'
```

src/s1reader/s1_orbit.py
```diff
@@ -255,6 +257,58 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
 
     # search for orbit file
     orbit_file_list = glob.glob(os.path.join(orbit_dir, 'S1*.EOF'))
+
+    orbit_file = get_orbit_file_from_list(zip_path, orbit_file_list)
+
+    if not orbit_file:
```
gshiroma, line 263:
```suggestion
    if orbit_file:
        return orbit_file
    if not auto_download:
        msg = (f'No orbit file was found for {os.path.basename(zip_path)} '
                f'from the directory provided: {orbit_dir}')
        warnings.warn(msg)
        return
```
Early return

src/s1reader/s1_orbit.py
```diff
@@ -255,6 +257,58 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
...
+        orbit_dir = os.path.dirname(orbit_dir)
+        if not orbit_dir:
```
gshiroma, line 265: `os.path.isdir(orbit_dir)` was tested above. What is the case that you are testing here?

src/s1reader/s1_orbit.py
```diff
@@ -255,6 +257,58 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
...
+            orbit_dir = os.getcwd()
+
+        if auto_download:
+            orbit_file = download_orbit(zip_path, orbit_dir)
+        else:
+            msg = (f'No orbit file found for {os.path.basename(zip_path)} '
+                    'from the orbit_source provided.')
+            warnings.warn(msg)
```
gshiroma, line 273:
```suggestion
    orbit_file = download_orbit(zip_path, orbit_dir)
```

src/s1reader/s1_orbit.py
```diff
@@ -255,6 +257,58 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
...
+
+    return orbit_file
+
+
+
+def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
+    '''Get orbit state vector list for a given swath.
```
gshiroma, line 280:
```suggestion
    '''Get orbit file for a given S-1 swath from a list of files
```

src/s1reader/s1_orbit.py
```diff
@@ -255,6 +257,58 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
...
+
+    Returns:
+    --------
+    orbit_file : str
+        Path to the orbit file.
+    '''
+
+    # check the existence of input file path and directory
+    #if not os.path.exists(zip_path):
+    #    raise FileNotFoundError(f"{zip_path} does not exist")
```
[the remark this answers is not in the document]
seongsujeong: Uncommenting the lines because some users would directly use `get_orbit_file_from_list()`

src/s1reader/s1_orbit.py
seongsujeong: Good catch. Agree that the test here is not necessary. Removing the lines here.

src/s1reader/s1_orbit.py
```diff
@@ -255,6 +257,58 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
...
+
+    #if not os.path.isdir(orbit_dir):
+    #    raise NotADirectoryError(f"{orbit_dir} not found")
+
+    # extract platform id, start and end times from swath file name
+    platform_id, t_swath_start_stop = get_file_name_tokens(zip_path)
+
+    # initiate output
+    orbit_file = ''
+
```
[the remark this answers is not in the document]
seongsujeong: The codes here is copied from `s1-orbit.py` in the current main branch. 

https://github.com/opera-adt/s1-reader/blob/4000bc7a31d79e740212ab31495ca28ad0e2d426/src/s1reader/s1_orbit.py#L254-L278

It should work okay as it has been, but also agree with your point. So I've made a little bit of changes. Please take a look at the changes I've committed.

src/s1reader/s1_orbit.py
seongsujeong: On the same context as above. Please see the changes I've made.

yunjunz (approved):
Thank you @seongsujeong. LGTM pending on the style issue flagged by Codacy.

gshiroma (approved):
LGTM! Thank you, @seongsujeong , for working on this and for addressing all my comments!

PR #80, opened 2022-11-12
Azimuth FM rate mismatch mitigation

seongsujeong:
This PR adds the functionality of calculating the azimuth FM rate mismatch mitigation described in ETAD ATBD (doc ID: ETAD-DLR-DD-0008), available [here](https://sentinels.copernicus.eu/documents/247904/4629150/Sentinel-1-ETAD-Product-Format-Specification.pdf)

Highlights of changes
- A new dataclass `BurstExtendedCoeffs` was added to `s1_annotation.py`. That is necessary to (linear) interpolate the FM rate and Doppler centroid along azimuth
- A new member `extended_coeffs_fm_dc` was added to `Sentinel1BurstSlc`
- A new method `az_fm_rate_mismatch_mitigation()` was added to `Sentinel1BurstSlc`, with auxiliary methods `_evaluate_polynomial_array()` and `_latlon_to_ecef()`

TODO:
~~- [ ] Replace `_latlon_to_ecef()` to equivalent ISCE3 function~~
- [x] Make sure it can take DEMs in arbitrary projections such as UTM and PS
- [x] Check if `isce3.io.Raster` can be generated on memory to avoid disk write


src/s1reader/s1_annotation.py
```diff
@@ -897,3 +951,115 @@ def _anx2height(cls, delta_anx):
             h_t += h_i * np.sin((i+1) * worb * delta_anx + phi[i])
 
         return h_t
+
+
+@dataclass
+class BurstExtendedCoeffs:
+    '''
+    Segments ofFM rate / Doppler centroid coefficients.
```
vbrancat, line 959: Mispelled. The docstring is not super clear to me. What does "Segments of FM rate" mean exactly?

src/s1reader/s1_annotation.py
```diff
@@ -897,3 +951,115 @@ def _anx2height(cls, delta_anx):
...
+        id_t0_fm_rate, id_t1_fm_rate = cls._find_t0_t1(product_annotation.vec_aztime_fm_rate,
+                                                       sensing_start,
+                                                       sensing_end)
+
+
+        # Scan the azimuth time of doppler centroid
+        id_t0_dc, id_t1_dc = cls._find_t0_t1(product_annotation.vec_aztime_dc,
+                                             sensing_start,
+                                             sensing_end)
+
```
vbrancat, line 1002: Remove white space

src/s1reader/s1_burst_slc.py
```diff
@@ -173,6 +175,10 @@ class Sentinel1BurstSlc:
     burst_noise: s1_annotation.BurstNoise  # Thermal noise correction
     burst_eap: s1_annotation.BurstEAP  # EAP correction
 
+    # Time series of FM rate / Doppler centroid polynomial coefficients
+    # for azimith FM rate mismatch mitigation
+    extended_coeffs: s1_annotation.BurstExtendedCoeffs
```
vbrancat, line 180: We do extrapolate the coefficients, but I think we just keep one. Do you need to extract all of them?

src/s1reader/s1_burst_slc.py
```diff
@@ -579,6 +585,318 @@ def az_carrier_components(self, offset, position):
...
+        eval_out = (arr_polynomial[:,0][..., np.newaxis] 
+                    *np.ones(ncol)[np.newaxis, ...]
+                    + (arr_polynomial[:,1][..., np.newaxis] 
+                       * np.ones(ncol)[np.newaxis, ...]) * term_tau
+                    + (arr_polynomial[:,2][..., np.newaxis]
+                       * np.ones(ncol)[np.newaxis, ...]) * term_tau**2)
+
+        return eval_out
+
+    def _llh_to_ecef(self, lat, lon, hgt, ellipsoid, unit_degree=True):
```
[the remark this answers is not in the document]
vbrancat: In which way does it look unconvenient? Can you elaborate on that?

src/s1reader/s1_burst_slc.py
```diff
@@ -579,6 +585,332 @@ def az_carrier_components(self, offset, position):
...
+            azimuth FM rate mismatch rate in radar grid in seconds.
+
+        '''
+        
+        os.makedirs(path_scratch, exist_ok=True)
+        
+        if custom_radargrid is None:
+            radargrid_correction = self.as_isce3_radargrid()
+        else:
+            radargrid_correction = custom_radargrid
```
vbrancat, line 696: Is this taking into account the possibility of generating the correction on a coarser radar grid?

src/s1reader/s1_burst_slc.py
```diff
@@ -579,6 +585,332 @@ def az_carrier_components(self, offset, position):
...
+
+        # Define the correction grid from the radargrid
+        # Also define the staggered grid in azimuth to calculate acceeration
+        width_grid = radargrid_correction.width
+        length_grid = radargrid_correction.length
+        intv_t = 1/radargrid_correction.prf
```
vbrancat, line 702:
```suggestion
        intv_t = 1.0/radargrid_correction.prf
```
Let's have it float

src/s1reader/s1_burst_slc.py
```diff
@@ -579,6 +585,332 @@ def az_carrier_components(self, offset, position):
...
+                           - intv_t / 2)
+
+        vec_tau = np.arange(width_grid)*intv_tau
+
+        grid_tau, grid_t = np.meshgrid(vec_tau, vec_t)
+
+        vec_position_intp = np.zeros((length_grid,3))
+        vec_vel_intp = np.zeros((length_grid,3))
+        vec_vel_intp_staggered = [None] * (length_grid+1)
+        for i_azimiuth, t_azimuth in enumerate(vec_t):
```
vbrancat, line 721:
```suggestion
        for i_azimuth, t_azimuth in enumerate(vec_t):
```
I am assuming the variable mispelling was not intentional

src/s1reader/s1_burst_slc.py
```diff
@@ -579,6 +585,332 @@ def az_carrier_components(self, offset, position):
...
+            vec_position_intp[i_azimiuth, :], vec_vel_intp[i_azimiuth, :] = \
+                self.orbit.interpolate(t_azimuth)
+
+        # Calculate velocity on staggered grid
+        # to kinematically calculate acceleration
+        for i_azimiuth, t_azimuth in enumerate(vec_t_staggered):
```
vbrancat, line 727:
```suggestion
        for i_azimuth, t_azimuth in enumerate(vec_t_staggered):
```
Same here?

src/s1reader/s1_burst_slc.py
```diff
@@ -579,6 +585,332 @@ def az_carrier_components(self, offset, position):
...
+
+        interpolator_tau0_fdc_interp = InterpolatedUnivariateSpline(
+                                        vec_aztime_dc_sec,
+                                        self.extended_coeffs.vec_tau0_dc,
+                                        k=1)
+        tau0_fdc_interp = interpolator_tau0_fdc_interp(vec_t)[..., np.newaxis]
+
+        # add range time origin to vec_tau
+        grid_tau += tau0_ka_interp * np.ones(vec_tau.shape)[np.newaxis, ...]
+
```
vbrancat, line 759:
```suggestion
```

src/s1reader/s1_burst_slc.py
```diff
@@ -579,6 +585,332 @@ def az_carrier_components(self, offset, position):
...
+        # Run topo on scratch directory
+        dem_raster = isce3.io.Raster(path_dem)
+        epsg = dem_raster.get_epsg()
+        proj = isce3.core.make_projection(epsg)
+        ellipsoid = proj.ellipsoid
+        grid_doppler = isce3.core.LUT2d()
+
+        Rdr2Geo = isce3.geometry.Rdr2Geo
+
+        # TODO Add parameters (# interation, threshold, etc.) to rdr2geo
```
vbrancat, line 789: I think this comment can be removed. I guess that the number of iterations and threshold will come from the COMPASS runconfig for geocoded CSLC?

src/s1reader/s1_burst_slc.py
```diff
@@ -579,6 +585,332 @@ def az_carrier_components(self, offset, position):
...
+        delta_t_freq_mm = (grid_freq_dc
+                           * (-1 / kappa_annotation_grid
+                              + 1 / kappa_annotation_true))
+
+        #return delta_t_freq_mm
+
+        # Prepare to export to LUT2d
+        vec_range = grid_tau(index_mid_burst_t) * isce3.speed_of_light / 2.0
+
+        return isce3.core.LUT2d(vec_range, vec_t, delta_t_freq_mm)
```
vbrancat, line 912: If I am not mistaken, for COMPASS, we have returned a LUT where the the first two vectors would define the grid on which the correction is computed in range and azimuth. We chose to express the vector in range using `meters` and the one in azimuth using `seconds`.

src/s1reader/s1_orbit.py
```diff
@@ -218,6 +218,8 @@ def get_file_name_tokens(zip_path: str) -> [str, list[datetime.datetime]]:
 # lambda to check if file exists if desired sat_id in basename
 item_valid = lambda item, sat_id: os.path.isfile(item) and sat_id in os.path.basename(item)
 
+
```
vbrancat, line 221:
```suggestion
```
Remove extra space

src/s1reader/s1_orbit.py
```diff
@@ -255,6 +257,53 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
...
+    if not auto_download:
+        msg = (f'No orbit file was found for {os.path.basename(zip_path)} '
+                f'from the directory provided: {orbit_dir}')
+        warnings.warn(msg)
+        return
+
+    # Attempt auto download
+    orbit_file = download_orbit(zip_path, orbit_dir)
+    return orbit_file
+
```
vbrancat, line 275:
```suggestion
```

src/s1reader/s1_orbit.py
```diff
@@ -255,6 +257,53 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
...
+
+
+def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
```
vbrancat, line 278: Are the changes related to `s1_orbit.py` actually caused by a merge with develop?

src/s1reader/s1_reader.py
```diff
@@ -649,6 +650,12 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
                                                                          aux_cal_subswath,
                                                                          sensing_start)
 
+        # Extenced FM and DC coefficient information
+        extended_coeffs = \
+            BurstExtendedCoeffs.from_product_annotation_and_burst(product_annotation,
```
vbrancat, line 655: Do we need to load the extended coefficient every time we are reading a burst? I am concerned that we are making the reader a bit slower reading metadata which might be not necessary for other workflows (e.g. RTC?)

vbrancat (commented):
Thank you @seongsujeong for this PR. Nice work.It is going to add a very crucial contribution to the CSLC development. I was able to follow the mathematical equations and compare them with the ETAD ATBD. I only have minor comments which I hope they will be addressed soon.

src/s1reader/s1_burst_slc.py
```diff
@@ -579,6 +585,332 @@ def az_carrier_components(self, offset, position):
...
+                        f'{path_scratch}/hgt_{str_datetime}.rdr']
+        
+        lat_raster, lon_raster, hgt_raster = [isce3.io.Raster(
+                                                        filename_raster,
+                                                        radargrid_correction.width,
+                                                        radargrid_correction.length,
+                                                        1,
+                                                        gdal.GDT_Float64,
+                                                        'ENVI')
+                                              for filename_raster in list_filename_llh]
```
LiangJYu, line 811:
```suggestion
        lat_raster, lon_raster, hgt_raster = [isce3.io.Raster(
                                                        f'{path_scratch}/{llh}_{str_datetime}.rdr'],
                                                        radargrid_correction.width,
                                                        radargrid_correction.length,
                                                        1,
                                                        gdal.GDT_Float64,
                                                        'ENVI')
                                              for llh in ['lat', 'lon', 'hgt']]
```
nit - fold in file name generation

src/s1reader/s1_burst_slc.py
```diff
@@ -579,6 +585,332 @@ def az_carrier_components(self, offset, position):
...
+            list_pos_satellite.append(vec_position_intp[:, i][..., np.newaxis]
+                                      * np.ones(grid_tau.shape[1])[np.newaxis, ...])
+            list_vel_satellite.append(vec_vel_intp[:, i][..., np.newaxis]
+                                      * np.ones(grid_tau.shape[1])[np.newaxis, ...])
+            list_acc_satellite.append(vec_acceleration_intp[:, i][..., np.newaxis]
+                                      * np.ones(grid_tau.shape[1])[np.newaxis, ...])
+
+        x_s, y_s, z_s = list_pos_satellite
+        vx_s, vy_s, vz_s = list_vel_satellite
+        ax_s, ay_s, az_s = list_acc_satellite
```
LiangJYu, line 889:
```suggestion
        tau_ones = np.ones(grid_tau.shape[1])[np.newaxis, ...]
        x_s, y_s, z_s = [vec_position_intp[:, i][..., np.newaxis] * tau_ones            
                         for i in range(3)]
        vx_s, vy_s, vz_s = [vec_vel_intp[:, i][..., np.newaxis] * tau_ones            
                            for i in range(3)]
        ax_s, ay_s, az_s = [vec_vel_intp[:, i][..., np.newaxis] * tau_ones            
                            for i in range(3)]
```
Brevity

src/s1reader/s1_burst_slc.py
```diff
@@ -579,6 +585,332 @@ def az_carrier_components(self, offset, position):
...
+
+        kappa_annotation_true = ( -(2 * self.radar_center_frequency)
+                                 / (isce3.core.speed_of_light * mag_xs_xg)
+                                 * (dotp_dxsg_acc + (vx_s**2 + vy_s**2 + vz_s**2)))
+
+        delta_t_freq_mm = (grid_freq_dc
+                           * (-1 / kappa_annotation_grid
+                              + 1 / kappa_annotation_true))
+
+        #return delta_t_freq_mm
```
LiangJYu, line 907:
```suggestion
```
Remove commented code

src/s1reader/s1_orbit.py
```diff
@@ -255,6 +257,53 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
...
+
+    orbit_file = get_orbit_file_from_list(zip_path, orbit_file_list)
+
+    if orbit_file:
+        return orbit_file
+
+    if not auto_download:
+        msg = (f'No orbit file was found for {os.path.basename(zip_path)} '
+                f'from the directory provided: {orbit_dir}')
+        warnings.warn(msg)
```
LiangJYu, line 269:
```suggestion
        raise FileNotFoundError(msg)
```
What do we think about raising an error instead of issuing a warning? An orbit is essential to processing no?

src/s1reader/s1_burst_slc.py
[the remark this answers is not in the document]
seongsujeong: Please correct me if I'm wrong, but It looks like `ellipsoid.lon_lat_to_xyz()` takes the a coordinate of a point.(i.e. (lon, lat, hgt). I need to convert the numpy array of longitude, latitude, and height into the array of x, y, and z (in ECEF).

src/s1reader/s1_annotation.py
seongsujeong: That `segments` here means the parts of the polynomial coefficient lists, along with the corresponding azimuth time in the annotation file. e.g. `generalAnnotation/azimuthFmRateList` or `dopplerCentroid/dcEstimateList`

We need to interpolate the polynomials along the azimuth, so I've extracted the minimum set of the polynomial information from the annotation.

src/s1reader/s1_burst_slc.py
seongsujeong: Yes. Also we can compare our calculation results with ETAD product by defining `custom_radargrid` in accordance with their product's correction grid.

src/s1reader/s1_burst_slc.py
seongsujeong: I think we are on the same understandings. The computation in the function `az_fm_rate_mismatch_mitigation()` is based on the grid whose unit in azimuth and range are seconds (by following the ETAD ATBD).
The line 910 above is to convert the range grid in time (tau) into meters, so that the return LUT2d is in compliance with our convention.

src/s1reader/s1_reader.py
seongsujeong: Good point Maybe we can add a flag to turn this on / off, like we did for EAP correction and AUX_CAL loader?

src/s1reader/s1_orbit.py
seongsujeong: Yes, especially the PR #79 to make the RTC workflow to work again.

src/s1reader/s1_burst_slc.py
seongsujeong: @vbrancat Can you show me the link to the examples where we do the extrapolation? That would be very helpful for me to  understand if that can be usable for the FM rate mismatch.

We need a sequence of the polynomials that is enough to cover the azimuth time of a burst. That is because we interpolate the polynomials along the azimuth time (of a burst), so usually 2~3 sets of polynomials are needed.

src/s1reader/s1_burst_slc.py
```diff
@@ -617,6 +623,313 @@ def az_carrier_components(self, offset, position):
...
+                                                        tau0_fdc_interp)
+
+        grid_freq_dc = (grid_freq_dcg
+                        + grid_kappa_t * ((grid_t - t_burst)
+                                          + (freq_dcg_burst / kappa_annotation_burst
+                                          - grid_freq_dcg / kappa_annotation_grid)))
+
+        lat_map = gdal.Open(list_filename_llh[0], gdal.GA_ReadOnly).ReadAsArray()
+        lon_map = gdal.Open(list_filename_llh[1], gdal.GA_ReadOnly).ReadAsArray()
+        hgt_map = gdal.Open(list_filename_llh[2], gdal.GA_ReadOnly).ReadAsArray()
```
seongsujeong, line 895: I initially tried the code using `with` as below:

```python
with (gdal.Open(list_filename_llh[0], gdal.GA_ReadOnly) as raster_lat,
      gdal.Open(list_filename_llh[1], gdal.GA_ReadOnly) as raster_lon,
      gdal.Open(list_filename_llh[2], gdal.GA_ReadOnly) as raster_hgt):
    lat_map = raster_lat.ReadAsArray()
    lon_map = raster_lon.ReadAsArray()
    hgt_map = raster_hgt.ReadAsArray()
```

which gave the error below:

```
Traceback (most recent call last):
  File "/Users/jeong/opera-adt/s1-reader_seongsujeong/src/s1reader/tester_s1_reader.py", line 37, in <module>
    azfmrmm = bursts[-1].az_fm_rate_mismatch_mitigation(
  File "/Users/jeong/miniconda3/envs/isce3_dev_rtc/lib/python3.9/site-packages/s1reader/s1_burst_slc.py", line 893, in az_fm_rate_mismatch_mitigation
    with (gdal.Open(list_filename_llh[0], gdal.GA_ReadOnly) as raster_lat,
AttributeError: __enter__
```

Maybe there is some incompatibility for GDAL raster with `with` statement (e.g. [here](https://stackoverflow.com/questions/49648360/use-with-statement-to-close-gdal-datasets) )

src/s1reader/s1_burst_slc.py
seongsujeong: Thanks @vbrancat for the snippets. I've revised the code to re-use the parsed coefficients (i.e. `az_fm_rate_list` and `doppler_list`) and extract the sequence that covers the burst's azimuth time.

src/s1reader/s1_burst_slc.py
seongsujeong: Yes. The task described in the TODO above should be addressed in `COMPASS` side, not `s1-reader`.

I'll submit a PR to COMPASS to integrate the azimuth FM mismatch mitigation into COMPASS.

src/s1reader/s1_reader.py
seongsujeong: Added a flag `flag_az_fmrate_mismatch_mitigation` to turn on / off the feature.

@vbrancat @LiangJYu Any suggestions for the flag name above? It looks too long to me.

src/s1reader/s1_burst_slc.py
```diff
@@ -579,6 +585,332 @@ def az_carrier_components(self, offset, position):
...
+        arr_coeff_b = np.array(list_b_interp).transpose()
+
+        # Run topo on scratch directory
+        dem_raster = isce3.io.Raster(path_dem)
+        epsg = dem_raster.get_epsg()
+        proj = isce3.core.make_projection(epsg)
+        ellipsoid = proj.ellipsoid
+        grid_doppler = isce3.core.LUT2d()
+
+        Rdr2Geo = isce3.geometry.Rdr2Geo
```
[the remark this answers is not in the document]
seongsujeong: Thanks for the suggestion. I've revised it to remove the alias, and to call `isce3.geometry.Rdr2Geo()` directly.

src/s1reader/s1_burst_slc.py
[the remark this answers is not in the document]
seongsujeong: @vbrancat Sorry that my response was not so clear. I was able to use the ISCE function to convert one coordinate, but not not sure how we can use it to convert ***array of*** coordinates.

For example let's say we have 2d numpy arrays `lon`, `lat`, and `hgt`.

This code works
```python
ellipsoid.lon_lat_to_xyz([lon[0, 0], lat[0, 0], hgt[0, 0]])
```

But this does not.
```python
ellipsoid.lon_lat_to_xyz([lon, lat, hgt])
```

@LiangJYu Do you know if there is some ISCE3 function that takes in numpy array for the LLH -> XYZ conversion?

src/s1reader/s1_orbit.py
seongsujeong: Makes sense to me, but just did not want to change the way that the current version of the function handles the missing orbit file (see below).
I'm open to changing the warning to Error. @vbrancat @hfattahi Any thoughts?

https://github.com/opera-adt/s1-reader/blob/d8f9e80364c02ef338062525065b1122599075d6/src/s1reader/s1_orbit.py#L266-L270

src/s1reader/s1_reader.py
```diff
@@ -470,7 +471,9 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
     return track_burst_num
 
 def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
-                   iw2_annotation_path: str, open_method=open, flag_apply_eap: bool = True):
+                   iw2_annotation_path: str, open_method=open,
+                   flag_apply_eap: bool = True,
+                   flag_az_fmrate_mismatch_mitigation = True):
```
LiangJYu, line 476:
```suggestion
                   mitigate_az_fmrate_mismatch = True):
```
nit - I do not like the verbosity of the `flag_` prefix. I think the variable names can be worded so the purpose of the variable is clear. `if mitigate_az_fmrate_mismatch:` reads as "if mitigate az fmrate mismatch then...".

src/s1reader/s1_annotation.py
```diff
@@ -897,3 +912,130 @@ def _anx2height(cls, delta_anx):
...
+            if (dt_t0 < 0) and (dt_t0_so_far < dt_t0):
+                id_t0_so_far = id_vec
+                dt_t0_so_far = dt_t0
+                continue
+
+            dt_t1 = (datetime_vec - datetime_end).total_seconds()
+            if (dt_t1 > 0) and (dt_t1_so_far > dt_t1):
+                id_t1_so_far = id_vec
+                dt_t1_so_far = dt_t1
+                continue
```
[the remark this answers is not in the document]
seongsujeong: @LiangJYu Thanks for the suggestion. Learning from your code here ;)

src/s1reader/s1_burst_slc.py
```diff
@@ -621,6 +626,312 @@ def az_carrier_components(self, offset, position):
...
+        Return:
+        -------
+        _: isce3.core.LUT2d
+            azimuth FM rate mismatch rate in radar grid in seconds.
+
+        '''
+
+        os.makedirs(path_scratch, exist_ok=True)
+
+        radargrid_correction = self.as_isce3_radargrid()
```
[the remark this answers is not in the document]
seongsujeong: I meant the radargrid (i.e. equivalent to "correction grid" in ETAD ATBD) on which the correction will be calculated.
What do you think about `correction_radargrid`?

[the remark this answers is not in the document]
LiangJYu: Sounds good to me

seongsujeong:
Regarding the LLH -> XYZ computation in the comment here: https://github.com/opera-adt/s1-reader/pull/80#discussion_r1036533764
Did some test to compare the performance. Looks like the numpy-based computation works several times faster than the ISCE3-based one.
Also checked the output values from the two methods, and there was no practical difference between them (<< 1.0e-9 meters).

So I will stick with the current implementation.

LiangJYu (approved):
LGTM

src/s1reader/s1_burst_slc.py
seongsujeong: @vbrancat Just to note, I've revised the function `az_fm_rate_mismatch_mitigation` to optionally take in range / azimuth steps, so that its interface looks more similar to `bistatic_delay` and `geometrical_and_steering_doppler`. Although, the azimuth FM rate calculation still needs DEM to calculate.

seongsujeong:
The figure below is the comparison of our calculation, the reference ETAD, and the difference (IW1). The jupyter notebook and the sample data is available upon request. Below are the detail about the test data.

SAFE file: S1A_IW_SLC__1SSV_20160805T230627_20160805T230654_012471_0137E2_B36E
Orbit file: S1A_OPER_AUX_POEORB_OPOD_20210313T055556_V20160804T225943_20160806T005943.EOF
DEM: Copernicus DEM, in EPSG3413 (cropped version of the .vrt version in the internal server)
ETAD: S1A_IW_ETA__AXSV_20160805T230627_20160805T230654_012471_0137E2_91C6

- Based on the information in the ETAD data, I reconstructed the radar grid as close as their correction grid
- The reconstructed radargrid was provided into the implemented function in this PR.
- We can see some terrain-like features in the difference plot (the 3rd column). Based on their pattern, I think that is caused by either difference in DEM resampling method, and/or difference in the starting time of the radargrid.



output

avalentino:
@seongsujeong ETAD provides both the radar grid times and the geodetic coordinates of each point in the grid itself. Have you compared that grids with yours?
Regarding the DEM, I thing that ETAD uses GLO-90 (DGED). Are you using the same?

src/s1reader/s1_burst_slc.py
```diff
@@ -106,6 +106,80 @@ class represents a polynomial function of range
...
+    hgt: np.ndarray
+        height as numpy array
+    ellipsoid: isce3.core.Ellipsoid
+        Definition of Ellipsoid
+    unit_degree: bool
+        True if the units of lat and lon are degrees.
+        False if the units are radian.
+
+    Return:
+    _: tuple(np.ndarray)
```
scottstanie, line 162:
```suggestion
    x_ecef : ndarray
        ECEF X coordinate (in meters).
    y_ecef : ndarray
        ECEF X coordinate (in meters).
    z_ecef : ndarray
        ECEF X coordinate (in meters).
```

src/s1reader/s1_burst_slc.py
```diff
@@ -106,6 +106,80 @@ class represents a polynomial function of range
...
+    -----------
+    lat: np.ndarray
+        latitude as numpy array
+    lon: np.ndarray
+        longitude as numpy array
+    hgt: np.ndarray
+        height as numpy array
+    ellipsoid: isce3.core.Ellipsoid
+        Definition of Ellipsoid
+    unit_degree: bool
```
scottstanie, line 157:
```suggestion
    in_degree: bool
```
(so that the check reads `if in_degrees:` rather than `if unit_degrees`)

src/s1reader/s1_burst_slc.py
```diff
@@ -621,6 +699,278 @@ def az_carrier_components(self, offset, position):
 
         return AzimuthCarrierComponents(kt, eta, eta_ref)
 
+    def az_fm_rate_mismatch_mitigation(self, path_dem: str,
+                                             path_scratch: str = None,
+                                             range_step=None,
+                                             az_step=None,
+                                             threshold_rdr2geo=1e-8,
+                                             numiter_rdr2geo=25,
+                                             custom_radargrid=None):
```
scottstanie, line 708:
```suggestion
    def az_fm_rate_mismatch_mitigation(self, 
                                       path_dem: str,
                                       path_scratch: str = None,
                                       range_step=None,
                                       az_step=None,
                                       threshold_rdr2geo=1e-8,
                                       numiter_rdr2geo=25,
                                       custom_radargrid=None):
```

src/s1reader/s1_burst_slc.py
```diff
@@ -621,6 +699,278 @@ def az_carrier_components(self, offset, position):
...
+        -----------
+        path_dem: str
+            Path to the DEM to calculate the actual azimuth FM rate
+        path_scratch: str
+            Path to the scratch directory to store intermediate data
+            If None, the scratch files will be saved on temporary directory generated by system
+        range_step: float
+            Range step of the correction grid in meters
+        az_step: float
+            Azimuth step of the correciton grid in seconds
```
scottstanie, line 725:
```suggestion
            Azimuth step of the correction grid in seconds
```

src/s1reader/s1_burst_slc.py
```diff
@@ -173,6 +247,10 @@ class Sentinel1BurstSlc:
     burst_noise: s1_annotation.BurstNoise  # Thermal noise correction
     burst_eap: s1_annotation.BurstEAP  # EAP correction
 
+    # Time series of FM rate / Doppler centroid polynomial coefficients
+    # for azimith FM rate mismatch mitigation
```
scottstanie, line 251:
```suggestion
    # for azimuth FM rate mismatch mitigation
```

src/s1reader/s1_reader.py
```diff
@@ -727,7 +743,8 @@ def _is_zip_annotation_xml(path: str, id_str: str) -> bool:
 
 def load_bursts(path: str, orbit_path: str, swath_num: int, pol: str = 'vv',
                 burst_ids: list[Union[str, S1BurstId]] = None,
-                flag_apply_eap: bool = True):
+                flag_apply_eap: bool = True,
+                mitigate_az_fmrate_mismatch = True):
```
scottstanie, line 747: I dont think this actually needs to be an argument here, or in `_burst_from_zip` or `_burst_from_safe_dir`:

The only thing this flag is turning on is whether to run `BurstExtendedCoeffs.from_polynomial_lists`. But there's no reason not to run that. This is a long and verbose flag just to calculate an extra set of polynomial coefficients and store them. The actual function is a user-run function from the Burst object, so it shouldn't matter whether they will possibly run it later when deciding if they are loading the bursts

scottstanie: The only change you'll need to do is 
at line 649:
1. make this 
```python
        if len(osv_list) > 0:
            # get orbit from state vector list/element tree
            sensing_duration = datetime.timedelta(
                seconds=n_lines * azimuth_time_interval)
            orbit = get_burst_orbit(sensing_start, sensing_start + sensing_duration,
                                    osv_list)
        else:
            orbit = None

```

into this

```python
        sensing_duration = datetime.timedelta(seconds=n_lines * azimuth_time_interval)
        if len(osv_list) > 0:
            # get orbit from state vector list/element tree
            orbit = get_burst_orbit(sensing_start, sensing_start + sensing_duration,
                                    osv_list)
        else:
            orbit = None
```
to avoid the undefined `sensing_duration` error if they don't pass an orbit (which I regularly do to get info about bursts)

src/s1reader/s1_reader.py
```diff
@@ -686,6 +691,16 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
 
+        # Extended FM and DC coefficient information
+        if mitigate_az_fmrate_mismatch:
+            extended_coeffs = \
+                BurstExtendedCoeffs.from_polynomial_lists(az_fm_rate_list,
+                                                        doppler_list,
+                                                        sensing_start,
+                                                        sensing_start + sensing_duration)
+        else:
+            extended_coeffs = None
```
scottstanie, line 702:
```suggestion
        # Extended FM and DC coefficient information
        extended_coeffs = BurstExtendedCoeffs.from_polynomial_lists(az_fm_rate_list,
                                                                    doppler_list,
                                                                    sensing_start,
                                                                    sensing_start + sensing_duration)
```

See comment at `load bursts` about running this function every time, no flag needed

src/s1reader/s1_burst_slc.py
```diff
@@ -621,6 +699,278 @@ def az_carrier_components(self, offset, position):
...
+            Maximum number of iteration for rdr2geo
+        custom_radargrid: isce3.product.RadarGridParameters
+            ISCE3 radar grid to define the correction grid.
+            If None, the full resolution radargrid of the burst will be used.
+
+        Return:
+        -------
+        _: isce3.core.LUT2d
+            azimuth FM rate mismatch rate in radar grid in seconds.
+
```
scottstanie, line 738: An `Examples` section could be good here so people know it's actually pretty easy to calculate, and that many of these options aren't necessary to get some answer:

```suggestion
    Examples
    ----------
    >>> correction_grid = burst.az_fm_rate_mismatch_mitigation("my_dem.tif")
```

scottstanie: Related- do we need to expose `numiter_rdr2geo`, `threshold_rdr2geo` to the user? will this be something we'll tinker with? It seems like someone may want to get corrections without knowing the implementation detail that we're using isce3's `rdr2geo` function to compute it.

scottstanie (approved):
I think most of my requests are maintainability-related changes, so i'll approved since you're trying to get this in quickly. 

The only one I'd definitely like to see done is the removal of the  `mitigate_az_fmrate_mismatch`, because its inclusion (and the default=True value) mean that `s1_info` breaks because you dont need an orbit to load bursts normally.

src/s1reader/s1_annotation.py
```diff
@@ -333,6 +335,19 @@ class ProductAnnotation(AnnotationBase):
...
+    # azimuth FM rate mismatch mitigation
+    vec_aztime_fm_rate: np.ndarray
+    lut_coeff_fm_rate: np.ndarray
+    vec_tau0_fm_rate:np.ndarray
+
+    # Doppler centroid (DC) parameters for
+    # azimuth FM rate mismatch mitigation
+    vec_aztime_dc: np.ndarray
+    lut_coeff_dc: np.ndarray
+    vec_tau0_dc: np.ndarray
```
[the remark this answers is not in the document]
seongsujeong: Good catch. There has been a change in the code during this PR, and those numpy arrays are not necessary anymore I will remove this line.

src/s1reader/s1_annotation.py
```diff
@@ -897,3 +912,120 @@ def _anx2height(cls, delta_anx):
...
+
+    @classmethod
+    def from_polynomial_lists(cls,
+                              az_fm_rate_list: list,
+                              doppler_centroid_list: list,
+                              sensing_start: datetime.datetime,
+                              sensing_end: datetime.datetime):
+        '''
+        Extract coefficients from `product_annotation` that fall within
```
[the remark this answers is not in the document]
seongsujeong: Looks like docstring needs to be revised (and I will). In the beginning of this PR, the lists of polynomials were stored in the data class `ProductAnnotation`, which was redundant because `s1_reader.burst_from_xml()` already loads the list of the polynomial coefficients.

src/s1reader/s1_burst_slc.py
```diff
@@ -8,7 +8,7 @@
 import isce3
 import numpy as np
 from osgeo import gdal
-
+from scipy.interpolate import InterpolatedUnivariateSpline
```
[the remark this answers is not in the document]
seongsujeong: Good point. The reason that I've used `InterpolatedUnivariateSpline` is because there is no guarantee that the grids (azimuth time in this case) are equally spaced. I'll do more search on this. For now, I'll add scipy into `requirements.txt` to ensure that we have everything we need.

src/s1reader/s1_burst_slc.py
```diff
@@ -106,6 +106,80 @@ class represents a polynomial function of range
...
+    eval_out = (arr_polynomial[:,0][..., np.newaxis]
+                *np.ones(ncol)[np.newaxis, ...]
+                + (arr_polynomial[:,1][..., np.newaxis]
+                    * np.ones(ncol)[np.newaxis, ...]) * term_tau
+                + (arr_polynomial[:,2][..., np.newaxis]
+                    * np.ones(ncol)[np.newaxis, ...]) * term_tau**2)
+
+    return eval_out
+
+def _llh_to_ecef(lat, lon, hgt, ellipsoid, unit_degree=True):
```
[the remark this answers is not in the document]
seongsujeong: I agree with your idea. This module is getting bigger and bigger, and would need some sort of refactoring. On top of sour suggestion, I think it would be necessary to move the correction algorithms (bistatic delay, EAP correction, FM rate mismatch, etc.) into separated code.

src/s1reader/s1_burst_slc.py
seongsujeong: > Related- do we need to expose `numiter_rdr2geo`, `threshold_rdr2geo` to the user? will this be something we'll tinker with? It seems like someone may want to get corrections without knowing the implementation detail that we're using isce3's `rdr2geo` function to compute it.

Makes sense. I think most of the users would not need to adjust those parameters. For them, just using the default values would work most of the case. But I would rather keep this as it is because it would not harm the function, and the function would have more freedom for those who are adventurous to play with the parameters.

src/s1reader/s1_annotation.py
```diff
@@ -897,3 +912,120 @@ def _anx2height(cls, delta_anx):
...
+
+    # FM rate
+    vec_aztime_fm_rate: np.ndarray
+    lut_coeff_fm_rate: np.ndarray
+    vec_tau0_fm_rate:np.ndarray
+
+    # Doppler centroid
+    vec_aztime_dc: np.ndarray
+    lut_coeff_dc: np.ndarray
+    vec_tau0_dc: np.ndarray
```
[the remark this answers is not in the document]
seongsujeong: 1. It might be redundant, but I would prefer to keep the prefix (or postfix) to explicitly indicate the shape of the variable (i.e. 1d or 2d array), so that helps the users to avoid the confusion about all the variables about fm_rate_* or dc_* going around the code. It helped me to write this code, so I believe the others should as well.
3. I agree that lut can be misunderstood as LUT object in ISCE3. I'll change `lut` to `arr`
4. Good idea to make the variable names start with `fm_rate_` and `dc_`. I will make the changes. I think that would also help improve the readability in the `1.`

src/s1reader/s1_reader.py
seongsujeong: > I dont think this actually needs to be an argument here, or in `_burst_from_zip` or `_burst_from_safe_dir`:
> 
> The only thing this flag is turning on is whether to run `BurstExtendedCoeffs.from_polynomial_lists`. But there's no reason not to run that. This is a long and verbose flag just to calculate an extra set of polynomial coefficients and store them. The actual function is a user-run function from the Burst object, so it shouldn't matter whether they will possibly run it later when deciding if they are loading the bursts

Agreed. I will make the "extra coefficient loader" to run every time.

PR #82, opened 2022-11-15
az time and slant corrections as functions of az time and slant range

LiangJYu:
Addresses comments from this [PR](https://github.com/opera-adt/COMPASS/pull/43) on az time and slant range correction LUTs.

### DIFF
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -404,7 +404,32 @@ class Sentinel1BurstSlc:
             self_as_dict[key] = val
         return self_as_dict
 
-    def bistatic_delay(self, xstep=1, ystep=1, use_radar_coords=True):
+
+    def _steps_to_aranges(self, range_step, az_step):
+        ''' convert range_step (meters) and az_step (seconds) into aranges to
+        generate LUT2ds
+        '''
+        n_range = np.ceil(self.width * self.range_pixel_spacing / range_step).astype(int)
+        range_arange = self.starting_range + np.arange(0, n_range) * range_step
+
+        n_az = np.ceil(self.length * self.azimuth_time_interval / az_step).astype(int)
+        rdrgrid = self.as_isce3_radargrid()
+        az_arange = rdrgrid.sensing_start + np.arange(0, n_az) * az_step
+
+        errs = []
+        if not range_arange:
+            errs.append('range')
+        if not az_arange:
+            errs.append('azimuth')
+        if errs:
+            errs = ', '.join(errs)
+            err_str = f'Cannot build aranges from following step(s): {errs}'
+            raise ValueError(err_str)
+
+        return range_arange, az_arange
+
+
+    def bistatic_delay(self, range_step=1, az_step=1):
         '''Computes the bistatic delay correction in azimuth direction
         due to the movement of the platform between pulse transmission and echo reception
         as described in equation (21) in Gisinger et al. (2021, TGRS).
@@ -421,10 +446,10 @@ class Sentinel1BurstSlc:
 
         Parameters
         -------
-        xstep : int
-           spacing along x direction (range direction) in units of pixels
-        ystep : int
-           spacing along y direction (azimuth direction) in units of pixels
+        range_step : int
+            Spacing along x/range direction [meters]
+        az_step : int
+            Spacing along y/azimuth direction [seconds]
         use_radar_coords : bool
             flag to construct correction as a function of azimuth time and
             slant range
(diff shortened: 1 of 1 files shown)


src/s1reader/s1_burst_slc.py
```diff
@@ -498,7 +506,8 @@ def geometrical_and_steering_doppler(self, xstep=500, ystep=50):
 
         return x, y, total_doppler
```
vbrancat, line 507: Don't we want to return this as a `isce3.core.LUT2d`? My understanding was that the `isce3.geocode.geocodeSlc `module is accepting the corrections as LUTs and not as `numpy` arrays.

vbrancat (commented):
Thank you @LiangJYu. I think we still have an issue with the way we provide the spacings for these two corrections. We are still assigning them in pixels but in COMPASS we decided to go for meters (range) and seconds (azimuth) respectively. I was hoping to get this solved in the `s1-reader`.

vbrancat:
What have we decided to do with the spacings of these two functions? Leave them in units of pixels but allow the CSLC runconfig in COMPASS to accept meters/seconds?

src/s1reader/s1_burst_slc.py
```diff
@@ -404,7 +404,32 @@ def as_dict(self):
...
+        az_arange = rdrgrid.sensing_start + np.arange(0, n_az) * az_step
+
+        errs = []
+        if not range_arange:
+            errs.append('range')
+        if not az_arange:
+            errs.append('azimuth')
+        if errs:
+            errs = ', '.join(errs)
+            err_str = f'Cannot build aranges from following step(s): {errs}'
```
scottstanie, line 426: Might be helpful to indicate why you can't build from the steps? 
Or rather, why would either of these fail? from a `self.width=0`? a failure with `self.as_isce3_radargrid().sensing_start`?

src/s1reader/s1_burst_slc.py
```diff
@@ -421,30 +446,29 @@ def bistatic_delay(self, xstep=1, ystep=1):
...
-        xstep : int
-           spacing along x direction (range direction) in units of pixels
-
-        ystep : int
-           spacing along y direction (azimuth direction) in units of pixels
+        range_step : int
+            Spacing along x/range direction [meters]
+        az_step : int
+            Spacing along y/azimuth direction [seconds]
+        use_radar_coords : bool
```
scottstanie, line 453: was this meant to be added as an option too? Right now there's just range/az step in the function

src/s1reader/s1_burst_slc.py
```diff
@@ -468,59 +492,66 @@ def geometrical_and_steering_doppler(self, xstep=500, ystep=50):
...
-           The index of samples in azimuth direction as an 1D array
-        total_doppler : float
-           Total Doppler which is the sum of the geometrical Doppler and
-           beam steering induced Doppler [Hz] as a 2D array
+           LUT2D object of total doppler in Hz as a function of the azimuth
+           time and slant range, or range and azimuth indices.
+           This correction needs to be added to the SLC tagged azimuth time to
+           get the corrected azimuth times.
         """
+        range_arange, az_arange = self._steps_to_aranges(range_step, az_step)
```
scottstanie, line 507: what do you think of the names `range_arr/az_arr` or `range_vec/az_vec`  to indicate they're arrays, rather than `_arange` (which says what function created them)?
Not a hard ask, just thinking that `range_arange` sounds a little funny haha

src/s1reader/s1_burst_slc.py
```diff
@@ -468,59 +492,66 @@ def geometrical_and_steering_doppler(self, xstep=500, ystep=50):
...
+        geometrical_doppler = self.doppler.poly1d.eval(range_arange)
 
         total_doppler = az_carr_comp.antenna_steering_doppler + geometrical_doppler
 
-        return x, y, total_doppler
+        return isce3.core.LUT2d(range_arange, az_arange, total_doppler)
 
-    def doppler_induced_range_shift(self, xstep=500, ystep=50):
+    def doppler_induced_range_shift(self, range_step=500, az_step=50,
+                                    use_radar_coords=True):
```
scottstanie, line 529: This parameter isn't used- where would it be passed to? Or should it be deleted?

scottstanie (approved):
This looks good to me except for the questions about deleting `use_radar_coords` references

src/s1reader/s1_burst_slc.py
LiangJYu: `width` and `length` should be ok as they're populated directly from the XML [here](https://github.com/opera-adt/s1-reader/blob/d948dc8215a9310eb7c80a496b0b8735028916e3/src/s1reader/s1_reader.py#L551-L552).

If `length * azimuth_time_interval < az_step` or `width * range_pixel_spacing < range_step`, `n_*` can be zero.

I will add check to ensure `*_step > 0`.

src/s1reader/s1_burst_slc.py
LiangJYu: I forgot to remove the above from the docstrings after modifying function.

src/s1reader/s1_burst_slc.py
LiangJYu: Agreed `_arange` is awkward. Will switch to `_vec`.

src/s1reader/s1_burst_slc.py
LiangJYu: Apologies for more slop. Thank you for catching this.

PR #87, opened 2022-11-21
fix LUT axis vector size check

LiangJYu:
Correctly check vector validity with `.size` property instead incorrectly treating it a python list/bool.


vbrancat (approved):
Thank you @LiangJYu, the fix worked!

seongsujeong (approved):
LGTM!

PR #92, opened 2022-12-14
FM rate mismatch - follow-up bug fix

seongsujeong:
It was found that some of the bursts' start-end time exceeded the range of azimuth time in polynomial data (FM rate, Doppler centroid). Such case seems to happen occasionally at the last burst in subswath. If the burst sensing time exceeds the polynomial list, then `BurstExtendedCoeffs.extract_polynomial_sequence()` will return empty array.

`Sentinel1BurstSlc.az_fm_rate_mismatch_mitigation()` will give an error when the polynomial information is empty.

This error would not cause RTC to crash, but will cause occasional error when calculating azimuth FM rate mismatch using the implementations included in the PR #80.

For the fix, I've added few lines to take care of that case.


vbrancat (approved):
LGTM! thanks for the fix

LiangJYu (approved):
LGTM

PR #100, opened 2023-02-13
s1_orbit.py: if autodownloading, create the directory

scottstanie:
There shouldnt be a need to have the user `mkdir -p orbits/` if they have already specified they do not have the orbits, and they want to download them.

This came up during COMPASS testing by @gracebato.

### DIFF
--- a/src/s1reader/s1_orbit.py
+++ b/src/s1reader/s1_orbit.py
@@ -246,7 +246,10 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
         raise FileNotFoundError(f"{zip_path} does not exist")
 
     if not os.path.isdir(orbit_dir):
-        raise NotADirectoryError(f"{orbit_dir} not found")
+        if not auto_download:
+            raise NotADirectoryError(f"{orbit_dir} not found")
+        else:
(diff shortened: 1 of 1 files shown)


vbrancat (approved):
Bumped also into this. Thanks for having fixed that @scottstanie

src/s1reader/s1_orbit.py
```diff
@@ -246,7 +246,10 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
         raise FileNotFoundError(f"{zip_path} does not exist")
 
     if not os.path.isdir(orbit_dir):
-        raise NotADirectoryError(f"{orbit_dir} not found")
+        if not auto_download:
+            raise NotADirectoryError(f"{orbit_dir} not found")
+        else:
+            os.makedirs(orbit_dir, exist_ok=True)
```
seongsujeong, line 252: What do you think about letting the user know that the directory has been created by printing some message to stdout?

seongsujeong (approved):
LGTM. I've left very minor comment in the below.

- Seongsu

src/s1reader/s1_orbit.py
scottstanie: i've added that print

PR #101, opened 2023-02-17
Two layered az fm mismatch rate

seongsujeong:
This PR is to separate the computation of azimuth FM rate mismatch mitigation (AZFMRMM hereinafter).
Currently, the function takes in the path to the DEM, run topo internally to get lon / lat / hgt in radar grid, and make use of them for the calculation. The new structure in this PR divides the procedure into two parts; The first part is a wrapper for the second part of the function. It takes in DEM, run topo, define the rear grid, passes them into the following function, and take the return object from the following function.
The second part is to take over what is calculated in the first part, and do the AZFMRMM calculation based on ETAD ATBD.

The reason for this separation is to reuse the topo outputs from the outside. Atmospheric corrections in COMPASS workflow makes use of incidence angle in radar grid, which can be obtained by running topo. Therefore, it would make more sense to run the topo once in COMPAS workflow, and pass the outputs to AZFMRMM (esp. to the second part) and the atmospheric corrections.


src/s1reader/s1_burst_slc.py
```diff
@@ -781,7 +787,93 @@ def az_fm_rate_mismatch_mitigation(self,
                                     length_radargrid,
                                     width_radargrid,
                                     self.as_isce3_radargrid().ref_epoch)
+        
+        # Run topo on scratch directory
+        dem_raster = isce3.io.Raster(path_dem)
```
vbrancat, line 792: I am wondering if it is more transparent for the users (and for testing) if we directly assign a DEM raster instead of a path. Also, we have no check at this stage that the DEM really exists. Probably, we should add one

src/s1reader/s1_burst_slc.py
```diff
@@ -781,7 +787,93 @@ def az_fm_rate_mismatch_mitigation(self,
...
+    def az_fm_rate_mismatch_from_llh(self,
+                                     lat_map: np.ndarray,
+                                     lon_map: np.ndarray,
+                                     hgt_map: np.ndarray,
+                                     ellipsoid: isce3.core.Ellipsoid,
+                                     correction_radargrid: isce3.product.RadarGridParameters):
+        '''
+        Take in lat / lon / hgt in radar grid, along with ellipsoid and the radar grid.
+        Calculate azimuth FM rate mismatch mitigation
+        Based on ETAD-DLR-DD-0008, Algorithm Technical Baseline Document.
```
vbrancat, line 853: I'd like to keep citation to references consistent with the numpy docstring. See Sec. 14 here https://numpydoc.readthedocs.io/en/latest/format.html

vbrancat (commented):
Thank you @seongsujeong for the quick fix. I still need to test it but for the code and the implementation I only have minor comments

vbrancat:
@LiangJYu can you please have a look at this? Thanks :)

LiangJYu:
> @LiangJYu can you please have a look at this? Thanks :)

I have reviewed already. My approval is pending the comment thread was regarding adding a DEM path check or switching to a DEM raster.

vbrancat:
@seongsujeong this PR is almost done. Let's make sure we implement the required changes soon and merge it asap

src/s1reader/s1_burst_slc.py
seongsujeong: @vbrancat I think it's a good idea to add a check at this point. @LiangJYu Thanks for the code suggestion.
On top of Liang's preference, in user perspective, I think it's simpler to provide the path to the DEM because we can put couple of routines (loading DEM as ISCE3 raster, file existence check) into one place.

LiangJYu (approved):
LGTM

vbrancat (approved):
LGTM, thanks @seongsujeong

PR #102, opened 2023-02-19
allow az_step/rg_step arguments to `as_isce3_radargrid`

scottstanie:
This allows a more coarse radar grid, as needed by some lookup tables such as the SET correction.

(this is mostly to help with cross-repo code duplication https://github.com/opera-adt/COMPASS/pull/91/files#diff-69dd9219c200228243e2fc9517100d1c8c2887874819ba403614548352b14b29R277-R292 )

### DIFF
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -282,26 +282,24 @@ class Sentinel1BurstSlc:
             RadarGridParameters constructed from class members.
         '''
 
+        width, length = self.shape
         if az_step is None:
             az_step = self.azimuth_time_interval
-            length = self.shape[0]
         else:
-            length_in_seconds = length / self.azimuth_time_interval
             if az_step < 0:
                 raise ValueError("az_step cannot be negative")
-            if 1 / az_step > length_in_seconds:
+            length_in_seconds = length * self.azimuth_time_interval
+            if az_step > length_in_seconds:
                 raise ValueError("az_step cannot be larger than radar grid")
-            length = (length / self.azimuth_time_interval) // az_step
-
+            length = length_in_seconds // az_step
 
         if rg_step is None:
             rg_step = self.range_pixel_spacing
-            width = self.shape[1]
         else:
-            width_in_meters = width * self.range_pixel_spacing
             if rg_step < 0:
                 raise ValueError("rg_step cannot be negative")
-            if rg_step > width * width_in_meters:
+            width_in_meters = width * self.range_pixel_spacing
+            if rg_step > width_in_meters:
                 raise ValueError("rg_step cannot be larger than radar grid")
             width = (width_in_meters) // rg_step
 
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -291,7 +291,7 @@ class Sentinel1BurstSlc:
             length_in_seconds = length * self.azimuth_time_interval
             if az_step > length_in_seconds:
                 raise ValueError("az_step cannot be larger than radar grid")
-            length = length_in_seconds // az_step
+            length = int(length_in_seconds / az_step)
 
         if rg_step is None:
             rg_step = self.range_pixel_spacing
@@ -301,7 +301,7 @@ class Sentinel1BurstSlc:
             width_in_meters = width * self.range_pixel_spacing
             if rg_step > width_in_meters:
                 raise ValueError("rg_step cannot be larger than radar grid")
-            width = (width_in_meters) // rg_step
+            width = int(width_in_meters / rg_step)
 
         prf = 1 / az_step
 
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -264,8 +264,16 @@ class Sentinel1BurstSlc:
     def __repr__(self):
         return f"{self.__class__.__name__}(burst_id={self.burst_id})"
 
-    def as_isce3_radargrid(self, az_step: Optional[float] = None, rg_step: Optional[float] = None):
+    def as_isce3_radargrid(self,
+                           az_step: Optional[float] = None,
+                           rg_step: Optional[float] = None):
         '''Init and return isce3.product.RadarGridParameters.
+
+        The `az_step` and `rg_step` parameters are used to construct a
+        decimated grid. If not specified, the grid will be at the full radar
+        resolution.
+        Note that increasing the range/azimuth step size does not change the sensing
+        start of the grid, as the grid is decimated rather than multilooked.
 
         Parameters
         ----------
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -2,6 +2,7 @@ import os
 from dataclasses import dataclass
 import datetime
 import tempfile
+from typing import Optional
 import warnings
 from packaging import version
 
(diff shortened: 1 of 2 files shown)


vbrancat (approved):
LGTM, thanks

src/s1reader/s1_burst_slc.py
```diff
@@ -263,17 +264,31 @@ def __str__(self):
...
+
+        Returns
+        -------
         _ : RadarGridParameters
             RadarGridParameters constructed from class members.
         '''
 
+        if az_step is None:
+            az_step = self.azimuth_time_interval
         prf = 1 / self.azimuth_time_interval
```
hfattahi, line 287: @scottstanie You need to also change the prf:

```suggestion
        prf = 1 / az_step
```

src/s1reader/s1_burst_slc.py
```diff
@@ -263,17 +264,31 @@ def __str__(self):
...
+            Range step size in meters. If not provided, the range step size
+            is set to the range pixel spacing.
+
+        Returns
+        -------
         _ : RadarGridParameters
             RadarGridParameters constructed from class members.
         '''
 
+        if az_step is None:
```
LiangJYu, line 285:
```suggestion
        length, width = self.shape
        if az_step is None:
            length_in_seconds = length / self.azimuth_time_interval
            if az_step < 0:
                raise ValueError("az_step cannot be negative")
            if 1 / az_step > length_in_seconds:
                raise ValueError("az_step cannot be larger than radar grid")
            length = (length / self.azimuth_time_interval) // az_step
```
Grid length will needs to be recalculated.
Also error checks for sanity. I think I have it right...
Move size extraction up to accommodate error checking.

src/s1reader/s1_burst_slc.py
```diff
@@ -263,17 +264,31 @@ def __str__(self):
...
+            az_step = self.azimuth_time_interval
         prf = 1 / self.azimuth_time_interval
 
+        if rg_step is None:
```
LiangJYu, line 289:
```suggestion        
        if rg_step is None:
            width_in_meters = width * self.range_pixel_spacing
            if rg_step < 0:
                raise ValueError("rg_step cannot be negative")
            if rg_step > width * width_in_meters:
                raise ValueError("rg_step cannot be larger than radar grid")
            width = (width_in_meters) // rg_step
```
Same treatment with range as above with azimuth.

src/s1reader/s1_burst_slc.py
scottstanie: I took the logic, but fixed a couple other things (like overwriting the length/width after this block)

src/s1reader/s1_burst_slc.py
```diff
@@ -263,18 +264,48 @@ def __str__(self):
...
         _ : RadarGridParameters
             RadarGridParameters constructed from class members.
         '''
 
-        prf = 1 / self.azimuth_time_interval
+        if az_step is None:
+            az_step = self.azimuth_time_interval
+            length = self.shape[0]
+        else:
+            length_in_seconds = length / self.azimuth_time_interval
```
hfattahi, line 289:
```suggestion
            length_in_seconds = length * self.azimuth_time_interval
```

src/s1reader/s1_burst_slc.py
```diff
@@ -263,18 +264,48 @@ def __str__(self):
...
+            if az_step < 0:
+                raise ValueError("az_step cannot be negative")
+            if 1 / az_step > length_in_seconds:
```
hfattahi, line 292:
```suggestion
            if  az_step > length_in_seconds:
```

src/s1reader/s1_burst_slc.py
```diff
@@ -263,18 +264,48 @@ def __str__(self):
...
+                raise ValueError("az_step cannot be larger than radar grid")
+            length = (length / self.azimuth_time_interval) // az_step
```
hfattahi, line 294:
```suggestion
            length = (length * self.azimuth_time_interval) // az_step
```

src/s1reader/s1_burst_slc.py
```diff
@@ -263,18 +264,48 @@ def __str__(self):
...
+
+
+        if rg_step is None:
+            rg_step = self.range_pixel_spacing
+            width = self.shape[1]
+        else:
+            width_in_meters = width * self.range_pixel_spacing
+            if rg_step < 0:
+                raise ValueError("rg_step cannot be negative")
+            if rg_step > width * width_in_meters:
```
hfattahi, line 304:
```suggestion
            if rg_step > width_in_meters:
```

hfattahi:
Also would you please add a test here: https://github.com/opera-adt/s1-reader/blob/main/tests/test_bursts.py

LiangJYu (approved):
LGTM

hfattahi:
Thank you @scottstanie for addressing the different comments and adding the test. 

Before approving this PR I would like to clarify one thing. 
Our ISCE3 convention is that in radar coordinates the "sensing start" and  "starting range" refer to the center of the first radar pixel. That's why if we multi-look, we have to account for the shift where the new center is. But if we decimate, that is slightly different and depending on the decimation convention we may keep the sensing start and starting range same as the original grid. In my opinion the changes in this PR is "decimation with zero offset" as defined [here](https://github.com/isce-framework/isce3/blob/develop/cxx/isce3/signal/decimate.h)

FYI, the radarGridParameters in isce3 already has a [multi-looking method](https://github.com/isce-framework/isce3/blob/develop/cxx/isce3/product/RadarGridParameters.h#L173).  I think we could just add a general decimation method there as well and these downstream changes wouldn't be needed. 

In the meantime if you all want to go ahead with this PR, I think I can approve but I suggest that we clarify in the doc string that:
```the input azimuth step size does not change the sensing start of the grid.``` and similarly for range

scottstanie:
Oh I didn't realize there was already the multilooking method! I wouldn't be against changing this to take `(az_looks, rg_looks)` instead, and just calling the isce3 method instead.

vbrancat:
> Oh I didn't realize there was already the multilooking method! I wouldn't be against changing this to take `(az_looks, rg_looks)` instead, and just calling the isce3 method instead.

I kind of find handy to directly assign the spacing of the radar grid :s

hfattahi:
> Oh I didn't realize there was already the multilooking method! I wouldn't be against changing this to take `(az_looks, rg_looks)` instead, and just calling the isce3 method instead.

But that's where exactly I wanted to clarify. What this PR meant to be? multi-looking or decimation? Do we want to change the sensing start, starting range or not?

vbrancat:
> > Oh I didn't realize there was already the multilooking method! I wouldn't be against changing this to take `(az_looks, rg_looks)` instead, and just calling the isce3 method instead.
> 
> But that's where exactly I wanted to clarify. What this PR meant to be? multi-looking or decimation? Do we want to change the sensing start, starting range or not?
> 
> This PR as is, is not same as multilooking method of the radar grid parameter.

@scottstanie was trying to simplify some code in my SET PR. We want to perform a pure decimation of the radar grid and keep the starting range and sensing start to be the same as the original radar grid. However, we do want a coarser grid for some of the corrections.

hfattahi:
> > > Oh I didn't realize there was already the multilooking method! I wouldn't be against changing this to take `(az_looks, rg_looks)` instead, and just calling the isce3 method instead.
> > 
> > 
> > But that's where exactly I wanted to clarify. What this PR meant to be? multi-looking or decimation? Do we want to change the sensing start, starting range or not?
> > This PR as is, is not same as multilooking method of the radar grid parameter.
> 
> @scottstanie was trying to simplify some code in my SET PR. We want to perform a pure decimation of the radar grid and keep the starting range and sensing start to be the same as the original radar grid. However, we do want a coarser grid for some of the corrections.

Thanks @vbrancat for the clarification. I chatted with @scottstanie offline. Let's move forward with this PR. But let's make the docstring as clear as it can be regarding the starting  range and azimuth time.

hfattahi (approved):
LGTM. Please clarify the docstring before merging as we discussed.

Issue #105, opened 2023-03-04
[Bug]: Inconsistent correction grid dimensions


vbrancat:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

Some correction grids within the s1-reader (e.g. bistatic delay and azimuth FM-rate mismatch) may have inconsistent dimensions.

For example the correction grid of the azimuth FM-rate mismatch may be computed using the decimation function of the radar grid reported here https://github.com/opera-adt/s1-reader/blob/0e0c7816a9b22b057a769519d302cd62e0a4e346/src/s1reader/s1_burst_slc.py#L267.

However, the correction grid for the bistatic delay is computed using the function reported here https://github.com/opera-adt/s1-reader/blob/0e0c7816a9b22b057a769519d302cd62e0a4e346/src/s1reader/s1_burst_slc.py#L531. The difference between the two function is using `np.ceil` when computing the dimension in both slant range and azimuth directions. 

This causes the two grid to be off by one pixel. For example, when the two grids are added together:

```
Traceback (most recent call last):
  File "/mnt/aurora-r0/vbrancat/codes/COMPASS/src/compass/s1_cslc.py", line 54, in <module>
    main()
  File "/mnt/aurora-r0/vbrancat/codes/COMPASS/src/compass/s1_cslc.py", line 49, in main
    run(parser.run_config_path, parser.args.grid_type)
  File "/mnt/aurora-r0/vbrancat/codes/COMPASS/src/compass/s1_cslc.py", line 43, in run
    s1_geocode_slc.run(cfg)
  File "/mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.11/site-packages/compass/s1_geocode_slc.py", line 82, in run
    rg_lut, az_lut = cumulative_correction_luts(burst,
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.11/site-packages/compass/utils/lut.py", line 60, in cumulative_correction_luts
    az_lut_data = -(bistatic_delay.data + az_fm_mismatch.data)
                    ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
ValueError: operands could not be broadcast together with shapes (110,408) (109,407) 
```

### What did you expect?

I expected the dimensions of all the correction grid contained in the repository to be consistent

PR #106, opened 2023-03-06
Consistent LUT grid shapes

vbrancat:
The azimuth LUT grids computed by the `s1-reader` may not have the same shape i.e., same number of lines and columns. This happens when the azimuth FM-rate mismatch is computed using rdr2geo layers (e.g. latitude, longitude, and height) coming from a decimated version of the radar grid. 

This PR fixs this inconsistency but computing the shape of the grid in the same way is done for the bistatic delay and the doppler. This PR should solve #105 .


vbrancat:
Closing because of duplicate of #107

PR #109, opened 2023-03-21
Add azimuth deramping function

vbrancat:
This PR introduce a functionality to compute the deramping signal function based on the carrier azimuth phase.

This functionality is needed for CSLC-S1 validation.


vbrancat:
We decided to output the ramp using the core module level. Therefore, this PR is not needed anymore

PR #111, opened 2023-04-07
support python 3.8 by using future.__annotations__

scottstanie:
only a few typing hints had `list`, so this change is rather minimal

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-python>=3.9
+python>=3.8
 numpy
 gdal
 packaging
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-python>=3.8
+python>=3.9
 numpy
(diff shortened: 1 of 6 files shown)


scottstanie:
Any reason for holding off on merging this?

gshiroma (approved):
LGTM. Thank you for this update, @scottstanie

seongsujeong (approved):
LGTM. I might not be in the conversation, but can you briefly explain why the support for python 3.8 was necessary?

scottstanie:
There wasn't any detailed conversation- it just seems like good practice to accommodate a python version that's not deprecated yet if there's no strong reason to move forward (like making heavy use of some feature that's not backward compatible). Doubly true for one like s1reader where python 3.8 incompatibility means all downstream packages can't use python 3.8 (like COMPASS, sweets, etc)

Issue #112, opened 2023-04-19
[Bug]: Azimuth FM rate mismatch fails for historical S1A data 


vbrancat:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

The azimuth FM-rate mismatch correction fails for some historical S1A data (2015-2017). The failure seems to be happening during the computation of the DC coefficients with the following error:

```
Traceback (most recent call last):
  File "/mnt/aurora-r0/vbrancat/codes/COMPASS/src/compass/s1_cslc.py", line 56, in <module>
    main()
  File "/mnt/aurora-r0/vbrancat/codes/COMPASS/src/compass/s1_cslc.py", line 51, in main
    run(parser.run_config_path, parser.args.grid_type)
  File "/mnt/aurora-r0/vbrancat/codes/COMPASS/src/compass/s1_cslc.py", line 43, in run
    s1_geocode_slc.run(cfg)
  File "/mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.11/site-packages/compass/s1_geocode_slc.py", line 77, in run
    cumulative_correction_luts(burst, dem_path=cfg.dem,
  File "/mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.11/site-packages/compass/utils/lut.py", line 65, in cumulative_correction_luts
    compute_geocoding_correction_luts(burst,
  File "/mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.11/site-packages/compass/utils/lut.py", line 214, in compute_geocoding_correction_luts
    az_fm_mismatch = burst.az_fm_rate_mismatch_from_llh(lat, lon, height,
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.11/site-packages/s1reader/s1_burst_slc.py", line 936, in az_fm_rate_mismatch_from_llh
    dc_coeffs = interp_coeffs(fm_rate_aztime_sec_vec,
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/aurora-r0/vbrancat/codes/miniconda3/envs/compass/lib/python3.11/site-packages/s1reader/s1_burst_slc.py", line 929, in interp_coeffs
[...]

PR #113, opened 2023-05-03
fix on DC polynomial interpolation for azimuth FM rate mismatch

seongsujeong:
This PR is to fix the bug reported in #112. The reason for the crash was because DC polynomial interpolator was using incorrect data in the time axis.
This fix was tested using the sample dataset reported in the issue above, and was able to give the LUT as below:

Screenshot 2023-05-03 at 10 31 17

Note on the tested dataset:
S1A_IW_SLC__1SSV_20150207T014953_20150207T015023_004511_005899_9C44.zip
S1A_OPER_AUX_POEORB_OPOD_20210305T194745_V20150206T225944_20150208T005944.EOF
Burst ID: `t064_135520_iw1`

### DIFF
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -931,7 +931,7 @@ class Sentinel1BurstSlc:
                 interpolated_coeffs.append(coeff_interpolator(az_time_interp))
             return np.array(interpolated_coeffs).transpose()
 
-        dc_coeffs = interp_coeffs(fm_rate_aztime_sec_vec,
+        dc_coeffs = interp_coeffs(dc_aztime_sec_vec,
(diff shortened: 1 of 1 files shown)


scottstanie:
Nice, I'll take a look at this soon.
Would we be able to add the SAFE file (even with just an empty .tiff) along with a new test to demo the fix?

seongsujeong:
@scottstanie Absolutely. The test data and the script are in the directory in aurora below:
`/u/aurora-r0/jeong/scratch/CSLC/historical_data`

Use `test_fmm.py` in the directory to run the demo.

vbrancat (approved):
LGTM, thanks for having fixed this

hfattahi (approved):

PR #119, opened 2023-05-28
Make the RFI data loading optional

scottstanie:
if we download only the annotation files via `asfsmd`, we want to be able to load bursts without needing the RFI

### DIFF
--- a/src/s1reader/s1_reader.py
+++ b/src/s1reader/s1_reader.py
@@ -5,8 +5,9 @@ import glob
 import os
 import warnings
 import lxml.etree as ET
-from typing import Union
 import zipfile
+from pathlib import Path
+from typing import Union
 
 from types import SimpleNamespace
 from packaging import version
@@ -569,11 +570,9 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
         tree_lads = ET.parse(f_lads)
         product_annotation = ProductAnnotation.from_et(tree_lads)
 
+        rfi_annotation_path = annotation_path.replace('annotation/', 'annotation/rfi/rfi-')
         # Load RFI information if available
-        if ipf_version >= RFI_INFO_AVAILABLE_FROM:
-            rfi_annotation_path =\
-                    annotation_path.replace('annotation/',
-                                            'annotation/rfi/rfi-')
+        if Path(rfi_annotation_path).exists() and ipf_version >= RFI_INFO_AVAILABLE_FROM:
             with open_method(rfi_annotation_path, 'r') as f_rads:
                 tree_rads = ET.parse(f_rads)
                 burst_rfi_info_swath = SwathRfiInfo.from_et(tree_rads,
--- a/src/s1reader/s1_reader.py
+++ b/src/s1reader/s1_reader.py
@@ -21,8 +21,7 @@ from s1reader import s1_annotation  # to access __file__
 from s1reader.s1_annotation import ProductAnnotation, NoiseAnnotation,\
                                    CalibrationAnnotation, AuxCal,\
                                    BurstCalibration, BurstEAP, BurstNoise,\
-                                   BurstExtendedCoeffs, SwathRfiInfo, SwathMiscMetadata,\
-                                   RFI_INFO_AVAILABLE_FROM
+                                   BurstExtendedCoeffs, SwathRfiInfo, SwathMiscMetadata
 
 from s1reader.s1_burst_slc import Doppler, Sentinel1BurstSlc
 from s1reader.s1_burst_id import S1BurstId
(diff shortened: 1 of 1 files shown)


src/s1reader/s1_reader.py
```diff
@@ -569,11 +570,9 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
         tree_lads = ET.parse(f_lads)
         product_annotation = ProductAnnotation.from_et(tree_lads)
 
+        rfi_annotation_path = annotation_path.replace('annotation/', 'annotation/rfi/rfi-')
         # Load RFI information if available
-        if ipf_version >= RFI_INFO_AVAILABLE_FROM:
-            rfi_annotation_path =\
-                    annotation_path.replace('annotation/',
-                                            'annotation/rfi/rfi-')
+        if Path(rfi_annotation_path).exists() and ipf_version >= RFI_INFO_AVAILABLE_FROM:
```
seongsujeong, line 575: I have a suggestion and a question:
- I think it would be a good idea to print out the warning message that RFI annotation was not found, so the reader is skipping RFI info loading.
- What's the difference between `os.path.exists(file_path)` and `Path(file_path).exists()`?

seongsujeong (commented):
LGTM overall. I've left a small comment about this PR.
In fact, this functionality is what I needed when testing the ANX time, which I temporarily placed the workaround locally on my end. Thank you for taking care of this @scottstanie

src/s1reader/s1_reader.py
scottstanie: 1. sure thing, I'll add that logging message
2. No difference- I've just started to find the `pathlib` API to be nicer to use, but i can change that if you want consistency

seongsujeong: Thanks for the response @scottstanie. In general I would like to suggest to import module as little as it can, even though it's included in python distribution.

>There should be one-- and preferably only one --obvious way to do it.
>  -- PEP20 - The Zen of Python

src/s1reader/s1_reader.py
```diff
@@ -569,18 +569,17 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
         tree_lads = ET.parse(f_lads)
         product_annotation = ProductAnnotation.from_et(tree_lads)
 
+        rfi_annotation_path = annotation_path.replace('annotation/', 'annotation/rfi/rfi-')
         # Load RFI information if available
-        if ipf_version >= RFI_INFO_AVAILABLE_FROM:
-            rfi_annotation_path =\
-                    annotation_path.replace('annotation/',
-                                            'annotation/rfi/rfi-')
+        if os.path.exists(rfi_annotation_path) and ipf_version >= RFI_INFO_AVAILABLE_FROM:
```
gshiroma, line 574: Considering that you are verifying the existence of the file, maybe you can remove `ipf_version >= RFI_INFO_AVAILABLE_FROM`? The only exception would be if there was some sort of RFI annotation prior to `RFI_INFO_AVAILABLE_FROM` with a different format

gshiroma (approved):
Changes look good to me. I have a minor comment below. Feel free to accept or reject the suggestion

src/s1reader/s1_reader.py
scottstanie: Good point, changed. I think in earlier products, the "product structure" being different just means the file doesn't exist.
image
https://sentinel.esa.int/documents/247904/1653442/DI-MPC-OTH-0540-1-0-RFI-Tech-Note.pdf/4b4fa95d-039f-5c78-fb90-06d307b3c13a?t=1644988601315

gshiroma: Awesome. Thank you, @scottstanie , for checking and updating the code

src/s1reader/s1_reader.py
seongsujeong: Interesting.
I'm not sure in what case we are going to have `/` at the end of the file path like what is in the example above. Can you  give me some example?

scottstanie: I was going this in Ipython where it autocompleted the file name, but in this case it was the SAFE directory (so it added a `'/'`). it was unexpected that a function to download the orbit would work/fail depending on if you include a trailing slash to the SAFE directory name

seongsujeong: I see. Interesting. That being case, I agree to switch to using `Path`

seongsujeong (approved):
LGTM. Thanks for the PR - Seongsu

scottstanie:
is this showing "merging is blocked" for other people? Not sure why it says that for me since there's 2 approvals now...

LiangJYu (approved):
LGTM

LiangJYu:
More reviews don't seem to help. Maybe it's the move isce framework GitHub group that's preventing PR from merging? IIRC we were in the OPERA-ADT GitHub group before. I can't think of anything else that's different.

scottstanie:
I invited Seongsu to the isce-framework org, apparently he wasn't a member 😿

seongsujeong:
@scottstanie Thanks for adding me to the organization :)

PR #120, opened 2023-05-30
estimate ascending node crossing from .EOF files

scottstanie:
Calculates the ANX crossing from a zip file and an EOF file. 

Right now
- it doesn't make the calculation from the POEORB the default ANX time. this is noted in the unit test, which will have to change
- im doing a dumb interpolation and we should probably just use `isce3.Orbit` instead of linear interp. of nearest two times
- we should, at the minimum, add the failing test case where it’s off by 3 seconds

for anyone taking it over, feel free to push to my branch (or take it and start your own)


scottstanie:
Implemented more fully in #121

Issue #117, opened 2023-05-24
`ESA burst map, burst database` != `Sentinel1BurstSlc.burst_id`


seongsujeong:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

It was discovered from the RTC result that some of the burst products has a cutoff. The issue took place especially around the coastal area of Brazil, Track number 126 in descending orbit. The QGIS screenshot below describes one of the examples.

<img width="2032" alt="Screenshot 2023-05-23 at 14 42 30" src="https://github.com/opera-adt/s1-reader/assets/27862199/999dd1d4-e8aa-498a-a503-29948f3b342b">

The yellow box is the coverage for `T126-269605-IW3` based on ESA burst map (and same for burst DB which is based on the burst map). The burst RTC products, which was for `T126-269605-IW3`, is right next to the yellow polygon. Based on the burst map, the RTC product above is covering the area for `T126-269604-IW3`. So I suspect that, for some reason, `T126-269604-IW3` was recognized as `T126-269605-IW3`, and the processing has pulled the bounding box info for `T126-269605-IW3`.

### What did you expect?

I think we need to solve the discrepancy in the burst map / database, and Sentinel1BurstSlc.burst_id

### Reproducible steps

```bash
More information about the example data shown above:
SAFE name: S1A_IW_SLC__1SDV_20210416T083427_20210416T083454_037473_046B05_5F0C
Burst ID: `t126-269605-IW3`
```

### Environment

```bash
- Version of this software [e.g. vX.Y.Z]
- Operating System: [e.g. MacOSX with Docker Desktop vX.Y]
...
```

seongsujeong:
More information that I've figured out so far
- Processed the data in 2019 and 2023. The bursts look okay in those cases:
  - S1A_IW_SLC__1SDV_20230112T083427_20230112T083454_046748_059ABA_08CC.zip
  - S1A_IW_SLC__1SDV_20190708T083419_20190708T083446_028023_032A31_3850.zip, S1A_IW_SLC__1SDV_20190708T083354_20190708T083421_028023_032A31_5F44.zip

hfattahi:
[...]

PR #121, opened 2023-06-08
Correction of ascending node crossing time  (ANX time)

seongsujeong:
It has came into our attention that some of Sentinel-1 data has noticeable error in ANX time, which can cause incorrect computation of burst ID. More details about the issue can be found in #117.
This PR is to fix the issue by calculating the ANX time from orbit data. The screenshot below is the RTC outputs (VV polarization) before and after the ANX time correction.

Screenshot 2023-06-07 at 17 04 24
Figure 1. Before the correction

.
Screenshot 2023-06-07 at 17 06 27
Figure 2. After the correction

Information about the test dataset above:
On and around [Grenada](https://en.wikipedia.org/wiki/Grenada) (North of Venezuela)
`S1B_IW_SLC__1SDV_20211014T095918_20211014T095945_029130_0379DF_B302.zip`
`S1B_OPER_AUX_POEORB_OPOD_20211103T112606_V20211013T225942_20211015T005942.EOF`

### DIFF
--- a/src/s1reader/s1_orbit.py
+++ b/src/s1reader/s1_orbit.py
@@ -13,7 +13,7 @@ from xml.etree import ElementTree
 FMT = "%Y%m%dT%H%M%S"
 
 # Orbital period of Sentinel-1
-T_ORBIT = (12*86400.0) / 175.0
+T_ORBIT = (12 * 86400.0) / 175.0
 
 # Temporal margin to apply to the start time of a frame
 #  - To make sure that the event of ascending node crossong is
--- a/src/s1reader/s1_orbit.py
+++ b/src/s1reader/s1_orbit.py
@@ -12,6 +12,10 @@ from xml.etree import ElementTree
 # date format used in file names
 FMT = "%Y%m%dT%H%M%S"
 
+# Temporal margin to apply to the start time of a frame - To make sure that ANX is included when choosing the orbit file
+MARGIN_START_FRAME_SECONDS = (12*86400.0) / 175.0 + 60.0
+margin_start_time = datetime.timedelta(seconds=MARGIN_START_FRAME_SECONDS)
+
 # Scihub guest credential
 scihub_user = 'gnssguest'
 scihub_password = 'gnssguest'
@@ -40,6 +44,9 @@ def download_orbit(safe_file: str, orbit_dir: str):
 
     # Parse info from SAFE file name
     mission_id, _, start_time, end_time, _ = parse_safe_filename(safe_file)
+
+    # Apply margin to the start time
+    start_time = start_time - margin_start_time
 
     # Find precise orbit first
     orbit_dict = get_orbit_dict(mission_id, start_time,
@@ -292,6 +299,9 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
 
     # extract platform id, start and end times from swath file name
     mission_id, t_swath_start_stop = get_file_name_tokens(zip_path)
+
+    # Apply temporal margin to the start time of the frame
+    t_swath_start_stop[0] = t_swath_start_stop[0] - margin_start_time
 
     # initiate output
     orbit_file_final = ''
--- a/src/s1reader/s1_orbit.py
+++ b/src/s1reader/s1_orbit.py
@@ -301,6 +301,7 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
     mission_id, t_swath_start_stop = get_file_name_tokens(zip_path)
 
     # Apply temporal margin to the start time of the frame
+    # 1st element: start time, 2nd element: end time
     t_swath_start_stop[0] = t_swath_start_stop[0] - margin_start_time
 
     # initiate output
--- a/src/s1reader/s1_orbit.py
+++ b/src/s1reader/s1_orbit.py
@@ -12,11 +12,12 @@ from xml.etree import ElementTree
 # date format used in file names
 FMT = "%Y%m%dT%H%M%S"
 
-# Orbital period of Sentinel-1
+# Orbital period of Sentinel-1 in seconds:
+# 12 days * 86400.0 seconds/day, divided into 175 orbits
 T_ORBIT = (12 * 86400.0) / 175.0
 
 # Temporal margin to apply to the start time of a frame
-#  - To make sure that the event of ascending node crossong is
+#  to make sure that the ascending node crossing is
 #    included when choosing the orbit file
 margin_start_time = datetime.timedelta(seconds=T_ORBIT + 60.0)
 
--- a/src/s1reader/s1_orbit.py
+++ b/src/s1reader/s1_orbit.py
@@ -12,9 +12,13 @@ from xml.etree import ElementTree
 # date format used in file names
 FMT = "%Y%m%dT%H%M%S"
 
-# Temporal margin to apply to the start time of a frame - To make sure that ANX is included when choosing the orbit file
-MARGIN_START_FRAME_SECONDS = (12*86400.0) / 175.0 + 60.0
-margin_start_time = datetime.timedelta(seconds=MARGIN_START_FRAME_SECONDS)
+# Orbital period of Sentinel-1
+T_ORBIT = (12*86400.0) / 175.0
+
+# Temporal margin to apply to the start time of a frame
+#  - To make sure that the event of ascending node crossong is
+#    included when choosing the orbit file
+margin_start_time = datetime.timedelta(seconds=T_ORBIT + 60.0)
 
 # Scihub guest credential
 scihub_user = 'gnssguest'
--- a/src/s1reader/s1_reader.py
+++ b/src/s1reader/s1_reader.py
@@ -532,7 +532,8 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
 
 def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
                                   sensing_time: datetime.datetime,
-                                  anx_time_annotation: datetime.datetime=None):
+                                  anx_time_annotation: datetime.datetime=None,
+                                  search_length=None):
     '''
     Estimate the time of ascending node crossing from orbit
 
@@ -552,6 +553,9 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
         - If it is `None`, then the orbit ANX time closest to (but no later than)
           the sensing time will be returned.
 
+    search_length: datetime.timedelta
+        Search length in time to crop the data in `orbit_state_vector_list`
+
     Returns
     -------
     _ : datetime.datetime
@@ -559,7 +563,8 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
     '''
 
     # Crop the orbit information before 2 * (orbit period) of sensing time
-    search_length = datetime.timedelta(seconds=2 * T_ORBIT)
+    if search_length is None:
+        search_length = datetime.timedelta(seconds=2 * T_ORBIT)
     orbit_until_sensing_time = get_burst_orbit(sensing_time  - search_length,
                                                sensing_time,
                                                orbit_state_vector_list)
@@ -813,7 +818,7 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
 
         # Calculate the difference in the two ANX times, and give
         # warning message when the difference is noticeable.
-        if not ascending_node_time_orbit is None:
+        if ascending_node_time_orbit is not None:
 
             diff_ascending_node_time_seconds = (
                 ascending_node_time_orbit - ascending_node_time_annotation).total_seconds()
--- a/src/s1reader/s1_reader.py
+++ b/src/s1reader/s1_reader.py
@@ -15,6 +15,7 @@ import isce3
 import numpy as np
 import shapely
 
+from scipy.interpolate import InterpolatedUnivariateSpline
 from nisar.workflows.stage_dem import check_dateline
 
 from s1reader import s1_annotation  # to access __file__
@@ -526,6 +527,45 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
 
     return track_burst_num
 
+
+def get_anx_time_orbit(osv_list:list, anx_time_annotation: datetime.datetime, margin_anx_sec=60.0):
+    '''
+    Estimate the time of ascending node crossing (ANX) from orbit
+
+    parameters
+    ----------
+    osv_list: list
+        Orbit state vectors as list
+    anx_time_annotatiion: datetime.datetime
+        ANX time read from annotation
+
+    returns
+    -------
+    _ : datetime.datetime
+        ANX time calculated from orbit
+    '''
+
+    margin_anx = datetime.timedelta(seconds=margin_anx_sec)
+    orbit_around_anx = get_burst_orbit(anx_time_annotation - margin_anx,
+                                       anx_time_annotation + margin_anx,
+                                       osv_list)
+    
+    grid_t_orbit = (np.arange(orbit_around_anx.time.size) * orbit_around_anx.time.spacing
+                    + orbit_around_anx.time.first)
+    grid_z_orbit = orbit_around_anx.position[:,2]
+    interpolator_time = InterpolatedUnivariateSpline(grid_z_orbit, grid_t_orbit, k=1)
+    
+    t_interp = interpolator_time(0.0)
+
+    datetime_ref = datetime.datetime(orbit_around_anx.reference_epoch.year,
+                                     orbit_around_anx.reference_epoch.month,
+                                     orbit_around_anx.reference_epoch.day)
+    datetime_ref += datetime.timedelta(
+        seconds=orbit_around_anx.reference_epoch.seconds_of_day())
+
+    return datetime_ref + datetime.timedelta(seconds=float(t_interp))
+
+
 def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
                    iw2_annotation_path: str, open_method=open,
                    flag_apply_eap: bool = True):
@@ -691,6 +731,14 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
     if orbit_path:
         orbit_tree = ET.parse(orbit_path)
         osv_list = orbit_tree.find('Data_Block/List_of_OSVs')
+
+        # Calculate ANX time from orbit; compare with the info from annotation
+        anx_time_orbit = get_anx_time_orbit(osv_list, ascending_node_time)
+        diff_anx_time_seconds = (anx_time_orbit - ascending_node_time).total_seconds()
+        if abs(diff_anx_time_seconds) > 1.0:
+            warnings.warn('Ascending node cross time is larger than 1.0 seconds. Using the time from orbit.')
+            ascending_node_time = anx_time_orbit
+
     else:
         osv_list = []
 
--- a/src/s1reader/s1_reader.py
+++ b/src/s1reader/s1_reader.py
@@ -28,7 +28,8 @@ from s1reader.s1_annotation import ProductAnnotation, NoiseAnnotation,\
 from s1reader.s1_burst_slc import Doppler, Sentinel1BurstSlc
 from s1reader.s1_burst_id import S1BurstId
 
-
+# Tolerance of the ascending node crossing (ANX) time
+ANX_TIME_TOLERANCE = 1.0
 esa_track_burst_id_file = f"{os.path.dirname(os.path.realpath(__file__))}/data/sentinel1_track_burst_id.txt"
 
 # TODO evaluate if it make sense to combine below into a class
@@ -730,17 +731,20 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
     # find orbit state vectors in 'Data_Block/List_of_OSVs'
     if orbit_path:
         orbit_tree = ET.parse(orbit_path)
-        osv_list = orbit_tree.find('Data_Block/List_of_OSVs')
+        orbit_state_vector_list = orbit_tree.find('Data_Block/List_of_OSVs')
 
-        # Calculate ascending node crossing (ANX) time from orbit; compare with the info from annotation
-        anx_time_orbit = get_anx_time_orbit(osv_list, ascending_node_time)
+        # Calculate ascending node crossing (ANX) time from orbit;
+        # compare with the info from annotation
+        anx_time_orbit = get_anx_time_orbit(orbit_state_vector_list, ascending_node_time)
         diff_anx_time_seconds = (anx_time_orbit - ascending_node_time).total_seconds()
-        if abs(diff_anx_time_seconds) > 1.0:
-            warnings.warn('Ascending node cross time is larger than 1.0 seconds. Using the time from orbit.')
+        if abs(diff_anx_time_seconds) > ANX_TIME_TOLERANCE:
+            warnings.warn('Ascending node cross time is larger than '
+                          f'{ANX_TIME_TOLERANCE} seconds. '
+                          'Using the time from orbit.')
             ascending_node_time = anx_time_orbit
 
     else:
-        osv_list = []
+        orbit_state_vector_list = []
 
     # load individual burst
     half_burst_in_seconds = 0.5 * (n_lines - 1) * azimuth_time_interval
@@ -772,11 +776,11 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
 
         sensing_duration = datetime.timedelta(
             seconds=n_lines * azimuth_time_interval)
-        if len(osv_list) > 0:
+        if len(orbit_state_vector_list) > 0:
             # get orbit from state vector list/element tree
 
             orbit = get_burst_orbit(sensing_start, sensing_start + sensing_duration,
-                                    osv_list)
+                                    orbit_state_vector_list)
         else:
             orbit = None
 
--- a/src/s1reader/s1_reader.py
+++ b/src/s1reader/s1_reader.py
@@ -530,7 +530,9 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
     return track_burst_num
 
 
-def get_ascending_node_time_orbit(orbit_state_vector_list: ET, sensing_time: datetime.datetime):
+def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
+                                  sensing_time: datetime.datetime,
+                                  anx_time_annotation: datetime.datetime=None):
     '''
     Estimate the time of ascending node crossing from orbit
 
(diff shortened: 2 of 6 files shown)


src/s1reader/s1_reader.py
```diff
@@ -526,6 +527,45 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
 
     return track_burst_num
 
+
+def get_anx_time_orbit(osv_list:list, anx_time_annotation: datetime.datetime, margin_anx_sec=60.0):
+    '''
+    Estimate the time of ascending node crossing (ANX) from orbit
+
+    parameters
```
gshiroma, line 535:
```suggestion
    Parameters
```

src/s1reader/s1_reader.py
```diff
@@ -526,6 +527,45 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
...
+    ----------
+    osv_list: list
+        Orbit state vectors as list
+    anx_time_annotatiion: datetime.datetime
+        ANX time read from annotation
+
+    returns
```
gshiroma, line 542:
```suggestion
    Returns
```

src/s1reader/s1_reader.py
```diff
@@ -526,6 +527,45 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
...
+
+def get_anx_time_orbit(osv_list:list, anx_time_annotation: datetime.datetime, margin_anx_sec=60.0):
+    '''
+    Estimate the time of ascending node crossing (ANX) from orbit
+
+    parameters
+    ----------
+    osv_list: list
+        Orbit state vectors as list
+    anx_time_annotatiion: datetime.datetime
```
gshiroma, line 539:
```suggestion
    anx_time_annotation: datetime.datetime
```

src/s1reader/s1_orbit.py
```diff
@@ -12,6 +12,10 @@
 # date format used in file names
 FMT = "%Y%m%dT%H%M%S"
 
+# Temporal margin to apply to the start time of a frame - To make sure that ANX is included when choosing the orbit file
+MARGIN_START_FRAME_SECONDS = (12*86400.0) / 175.0 + 60.0
```
gshiroma, line 16: Where did these numbers came from?

src/s1reader/s1_orbit.py
```diff
@@ -293,6 +300,9 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
     # extract platform id, start and end times from swath file name
     mission_id, t_swath_start_stop = get_file_name_tokens(zip_path)
 
+    # Apply temporal margin to the start time of the frame
+    t_swath_start_stop[0] = t_swath_start_stop[0] - margin_start_time
```
gshiroma, line 304: The comment says "to the start time". Should it be "to the stop time"?

src/s1reader/s1_reader.py
```diff
@@ -526,6 +527,45 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
...
     return track_burst_num
 
+
+def get_anx_time_orbit(osv_list:list, anx_time_annotation: datetime.datetime, margin_anx_sec=60.0):
+    '''
+    Estimate the time of ascending node crossing (ANX) from orbit
+
+    parameters
+    ----------
+    osv_list: list
```
gshiroma, line 537:
```suggestion
    osv_list: list
```
Please, consider using a variable name that is easier to understand and read, e.g. 'state_vector_list'. Also, please describe briefly how each state vector should be describe. Is it a tuple/list of 2 or 3 elements?

gshiroma (commented):
Thank you, @seongsujeong , for working on that. I have a few comments below more related to "style", but I'd like to understand what you are doing here. Could you please provide more description in the PR and in the code? We can also schedule a meeting so you can explain this PR so we can understand it faster.

src/s1reader/s1_orbit.py
seongsujeong: TODO: clarify what the numbers in the equations are for

src/s1reader/s1_orbit.py
```diff
@@ -293,6 +300,9 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
     # extract platform id, start and end times from swath file name
     mission_id, t_swath_start_stop = get_file_name_tokens(zip_path)
 
+    # Apply temporal margin to the start time of the frame
```
seongsujeong, line 303:
```suggestion
    # Apply temporal margin to the start time of the frame
    # 0th elem: start time, 1st elem: end time
```

src/s1reader/s1_reader.py
seongsujeong: TODO: change the variable name throughout the code

src/s1reader/s1_reader.py
```diff
@@ -526,6 +527,45 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
...
+        Orbit state vectors as list
+    anx_time_annotatiion: datetime.datetime
+        ANX time read from annotation
+
+    returns
+    -------
+    _ : datetime.datetime
+        ANX time calculated from orbit
```
seongsujeong, line 545:
```suggestion
        Ascending node crossing (ANX) time calculated from orbit
```

src/s1reader/s1_reader.py
```diff
@@ -526,6 +527,45 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
...
+def get_anx_time_orbit(osv_list:list, anx_time_annotation: datetime.datetime, margin_anx_sec=60.0):
+    '''
+    Estimate the time of ascending node crossing (ANX) from orbit
+
+    parameters
+    ----------
+    osv_list: list
+        Orbit state vectors as list
+    anx_time_annotatiion: datetime.datetime
+        ANX time read from annotation
```
seongsujeong, line 540:
```suggestion
        ANX time estimated from annotation
```

src/s1reader/s1_reader.py
```diff
@@ -691,6 +731,14 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
     if orbit_path:
         orbit_tree = ET.parse(orbit_path)
         osv_list = orbit_tree.find('Data_Block/List_of_OSVs')
+
+        # Calculate ANX time from orbit; compare with the info from annotation
```
seongsujeong, line 735:
```suggestion
        # Calculate ascending node crossing (ANX) time from orbit; compare with the info from annotation
```

src/s1reader/s1_reader.py
```diff
@@ -691,6 +731,14 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
+        anx_time_orbit = get_anx_time_orbit(osv_list, ascending_node_time)
+        diff_anx_time_seconds = (anx_time_orbit - ascending_node_time).total_seconds()
+        if abs(diff_anx_time_seconds) > 1.0:
```
seongsujeong, line 738: Change the hard-coded `1.0` to a constant in the code. Do more a bit of testing about how sensitive the Burst ID computation is w.r.t. the error in ANX time.

src/s1reader/s1_orbit.py
seongsujeong: It's the orbital period of Sentinel-1 + 60 seconds. When searching for the orbit file for SAFE file, this is added to the sensing time of the frame data to make sure that the event of ascending node cross is included in the orbit file search result.

src/s1reader/s1_orbit.py
seongsujeong: Explained in the meeting, but also leaving here for the record. `t_swath_start_stop` is a list of `datetime.datetime` with two elements. The first and seconds elements in the list is starting and end time respectively.

src/s1reader/s1_reader.py
seongsujeong: The variable name `osv_list` was changed to `orbit_state_vector_list` throughout the code.

src/s1reader/s1_orbit.py
```diff
@@ -12,6 +12,14 @@
 # date format used in file names
 FMT = "%Y%m%dT%H%M%S"
 
+# Orbital period of Sentinel-1
+T_ORBIT = (12 * 86400.0) / 175.0
```
gshiroma, line 16:
```suggestion
# Orbital period of Sentinel-1 in seconds:
# 12 days * 86400.0 seconds/day, divided into 175 orbits
T_ORBIT = (12 * 86400.0) / 175.0
```

src/s1reader/s1_orbit.py
```diff
@@ -12,6 +12,14 @@
...
+
+# Temporal margin to apply to the start time of a frame
+#  - To make sure that the event of ascending node crossong is
```
gshiroma, line 19:
```suggestion
#  to make sure that the ascending node crossing is
```

src/s1reader/s1_reader.py
```diff
@@ -26,8 +27,9 @@
 
 from s1reader.s1_burst_slc import Doppler, Sentinel1BurstSlc
 from s1reader.s1_burst_id import S1BurstId
-
-
+from s1reader.s1_orbit import T_ORBIT
```
gshiroma, line 30:
```suggestion
from s1reader.s1_orbit import T_ORBIT

```

src/s1reader/s1_reader.py
```diff
@@ -26,8 +27,9 @@
...
+# Tolerance of the ascending node crossing time
```
gshiroma, line 31:
```suggestion
# Tolerance of the ascending node crossing time in seconds
```

src/s1reader/s1_reader.py
```diff
@@ -526,6 +528,80 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
...
+
+def get_ascending_node_time_orbit(orbit_state_vector_list: ET, sensing_time: datetime.datetime):
+    '''
+    Estimate the time of ascending node crossing from orbit
+
+    Parameters
+    ----------
+    orbit_state_vector_list: ET
+        XML elements that points to the list of orbit information.
+        each elements in the parameter contains the information below:
```
gshiroma, line 540:
```suggestion
        Each element should contain the information below:
```

src/s1reader/s1_reader.py
```diff
@@ -526,6 +528,80 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
...
+        Ascending node crossing time calculated from orbit
+    '''
+
+    # Crop the orbit information before 2 * (orbit period) of sensing time
+    search_length = datetime.timedelta(seconds=2 * T_ORBIT)
+    orbit_until_sensing_time = get_burst_orbit(sensing_time  - search_length,
+                                               sensing_time,
+                                               orbit_state_vector_list)
+
+    # Convert ISCE3 oribt object's reference datetime into python object
```
gshiroma, line 558:
```suggestion
    # Convert the ISCE3 orbit reference datetime into a python object
```

src/s1reader/s1_reader.py
```diff
@@ -526,6 +528,80 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
...
+    datetime_orbit_ref = datetime.datetime(
+        orbit_until_sensing_time.reference_epoch.year,
+        orbit_until_sensing_time.reference_epoch.month,
+        orbit_until_sensing_time.reference_epoch.day)
+    datetime_orbit_ref += datetime.timedelta(
+        seconds=orbit_until_sensing_time.reference_epoch.seconds_of_day())
+
+    # detect the event of ascending node crossing from the cropped orbit info.
+    # Algorithm inspirates by Scott Staniewicz's PR in the link below:
```
gshiroma, line 567:
```suggestion
    # Detect the event of ascending node crossing from the cropped orbit info.
    # The algorithm was inspired by Scott Staniewicz's PR in the link below:
```

src/s1reader/s1_reader.py
```diff
@@ -526,6 +528,80 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
...
+                                                             k=1)
+            t_interp = interpolator_time(0.0)
+
+            # Convert the interpolated time into datetime
+            # Add the ascending node crossing time if it's before the sensing time
+            datetime_ascending_crossing = (datetime_orbit_ref
+                                           + datetime.timedelta(seconds=float(t_interp)))
+            if datetime_ascending_crossing < sensing_time:
+                datetime_ascending_node_crossing_list.append(datetime_ascending_crossing)
+        z_prev = z
```
gshiroma, line 596:
```suggestion
        if z_prev is None or not (z_prev < 0 <= z):
            z_prev = z
            continue
        # Extract z coords. and time for interpolation
        index_from = max(index_z - 3, 0)
        index_to = min(index_z + 3, len(orbit_z_vec))
        z_around_crossing = orbit_z_vec[index_from : index_to]
        time_around_crossing = orbit_time_vec[index_from : index_to]

        # Set up spline interpolator and interpolate the time when z is equal to 0.0
        interpolator_time = InterpolatedUnivariateSpline(z_around_crossing,
                                                         time_around_crossing,
                                                         k=1)
        t_interp = interpolator_time(0.0)

        # Convert the interpolated time into datetime
        # Add the ascending node crossing time if it's before the sensing time
        datetime_ascending_crossing = (datetime_orbit_ref
                                       + datetime.timedelta(seconds=float(t_interp)))
        if datetime_ascending_crossing < sensing_time:
            datetime_ascending_node_crossing_list.append(datetime_ascending_crossing)

```

src/s1reader/s1_reader.py
```diff
@@ -26,8 +27,9 @@
 
 from s1reader.s1_burst_slc import Doppler, Sentinel1BurstSlc
 from s1reader.s1_burst_id import S1BurstId
-
-
+from s1reader.s1_orbit import T_ORBIT
+# Tolerance of the ascending node crossing time
+ASCENDING_NODE_TIME_TOLERANCE = 1.0
```
gshiroma, line 32: How about renaming it to `ASCENDING_NODE_TIME_TOLERANCE_IN_SEC`?

gshiroma (commented):
Thank you, @seongsujeong , for addressing my comments. I have some other minor comments below.

src/s1reader/s1_reader.py
```diff
@@ -682,17 +766,42 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
+            # Calculate the difference in the times.
+            # Give warning message when the difference is noticeable.
+            diff_ascending_node_time_seconds = (
+                ascending_node_time_orbit - ascending_node_time_annotation).total_seconds()
+            if abs(diff_ascending_node_time_seconds) > ASCENDING_NODE_TIME_TOLERANCE_IN_SEC:
+                warnings.warn('ascending node time error larger than '
+                            f'{ASCENDING_NODE_TIME_TOLERANCE_IN_SEC} seconds was detected: '
+                            f'Error = {diff_ascending_node_time_seconds} seconds.')
+
+            ascending_node_time = ascending_node_time_orbit
```
gshiroma, line 798: This assignment `ascending_node_time = ascending_node_time_orbit` is outside the if statement `if abs(diff_ascending_node_time_seconds) > ASCENDING_NODE_TIME_TOLERANCE`, so it will happen for cases where the ANX time difference is less than or equal `ASCENDING_NODE_TIME_TOLERANCE`. 

Is that what you intended? I thought you'd only use `ascending_node_time_orbit` if the ANX time difference was greater than the tolerance.

src/s1reader/s1_reader.py
seongsujeong: I  understand this suggestion as to the improvement of the variable initialization and the readability improvement. I've taken another approach to achieve this.

src/s1reader/s1_reader.py
```diff
@@ -682,17 +766,42 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
+        orbit_state_vector_list = orbit_tree.find('Data_Block/List_of_OSVs')
+
+        # Calculate ascending node crossing time from orbit;
+        # compare with the info from annotation
+        ascending_node_time_orbit = get_ascending_node_time_orbit(
+            orbit_state_vector_list, first_line_utc_time)
+        if ascending_node_time_orbit is None:
+            warnings.warn('Cannot estimate ascending node crossing time from Orbit. '
+                          'Using the ascending node time in annotation.')
+            ascending_node_time = ascending_node_time_annotation
```
scottstanie, line 787: What if we made `get_ascending_node_time_orbit` raise a ValueError with `Cannot estimate ascending node crossing time from Orbit.` instead of returning `None`? It seems like from the perspective of that function, it would be an error to pass an invalid orbit file/the wrong orbit file.
Then this part of the code would be a `try/except` instead of checking `if ascending_node_time_orbit is None:`

src/s1reader/s1_reader.py
seongsujeong: Early return has implemented as discussed in the RTC meeting :)

src/s1reader/s1_reader.py
seongsujeong: For the record:
- By default, the ANX time will be estimated from the orbit
- ANX time from the annotation will be used if:
  - The orbit file was not provided into the reader, or
  - ANX time estimation from the orbit is not successful.

The snippet above is to let the user informed that there are significant difference between the two ANX times. I thought it would be useful for the users about what is happening with the data.

gshiroma (approved):
Thank you, @seongsujeong , for implementing this fix and for addressing all my reviews.

scottstanie:
@seongsujeong would you be able to pull in the unit test from here? https://github.com/opera-adt/s1-reader/pull/120/files#diff-9eb0a8ba9a467c950451e7fe6228094f7511aff2d0fe6a423c0471d7c8b3de3f (i'll close that PR now)

scottstanie (changes requested):
Looks good, with the main necessary change being the unit test addition 👍

seongsujeong:
Sorry for the delay. The updates in the ANX time detection has discovered some cases where the ANX time is one T_orb behind of what it should be (i.e. ANX time in the previous orbit). There were 12k - 13k frames per year, and S1A seems to have more cases than S1B.

The existing code in `s1_burst_id.py` below seems to take of such cases already (nice work @scottstanie ), but I'm testing the new code in this PR to make sure.

https://github.com/opera-adt/s1-reader/blob/ed16185338c879b59518ca26936751a1baffd621/src/s1reader/s1_burst_id.py#L90-L98

seongsujeong:
@hfattahi  @scottstanie  @gshiroma @vbrancat Opening the discussion here for more open discussion.

There is an edge case that current PR about the ascending node crossing (ANX) time causes an unexpected result.
This issue took place about 50 times over ~7000 tested SAFE frames whose `[sensing time - ANX time]` (i.e. `azimuthAnxTime`) is longer than the orbit period (`T_orb`, ~98 .7minutes).

Below are the information about the example cases:

SAFE: `S1A_IW_SLC__1SDV_20200102T111719_20200102T111747_030620_038227_9FE9`
ORBIT: `S1A_OPER_AUX_POEORB_OPOD_20210316T184157_V20200101T225942_20200103T005942.EOF`

#### Symptom:
The reader before this PR gives correct burst IDs which was confirmed by the comparison between the ESA burst map.
In contrast, the current fix in this PR ended up with incorrect burst IDs. For example, one of the bursts `t099_210499_iw1` in the SAFE frame above has resulted in `t098_208351_iw1` with this PR. Below is the full list of burst IDs from the main branch, and from this PR


from main branch (in agreement with burst map):
>t099_210499_iw1, t099_210500_iw1, t099_210501_iw1, t099_210502_iw1, t099_210503_iw1, t099_210504_iw1, t099_210505_iw1, t099_210506_iw1, t099_210507_iw1, t099_210498_iw2, t099_210499_iw2, t099_210500_iw2, t099_210501_iw2, t099_210502_iw2, t099_210503_iw2, t099_210504_iw2, t099_210505_iw2, t099_210506_iw2, t099_210507_iw2, t099_210498_iw3, t099_210499_iw3, t099_210500_iw3, t099_210501_iw3, t099_210502_iw3, t099_210503_iw3, t099_210504_iw3, t099_210505_iw3, t099_210506_iw3

from this PR (NOT in agreement with burst map):
> t098_208351_iw1, t098_208352_iw1, t098_208353_iw1, t098_208354_iw1, t098_208355_iw1, t098_208356_iw1, t098_208357_iw1, t098_208358_iw1, t098_208359_iw1, t098_208350_iw2, t098_208351_iw2, t098_208352_iw2, t098_208353_iw2, t098_208354_iw2, t098_208355_iw2, t098_208356_iw2, t098_208357_iw2, t098_208358_iw2, t098_208359_iw2, t098_208350_iw3, t098_208351_iw3, t098_208352_iw3, t098_208353_iw3, t098_208354_iw3, t098_208355_iw3, t098_208356_iw3, t098_208357_iw3, t098_208358_iw3



#### Cause:
`manifest.safe` of the SAFE frame tells that the start / stop relative orbit of the frame is `98` and `99` respectively. The absolute orbit in the manifest was `30620` and `30621`respectively.
However, in fact, the frame did not have any burst from track 98.

Screenshot 2023-07-17 at 13 38 36
Big red box with empty filling is the coverage of the SAFE frame above. All bursts in the frame had track number `99`.

The smaller Red box right below the SAFE frame was the last burst in track 98 Its burst ID is shown in the right panel in the screenshot.

Red solid line in the below of those boxes is equator.

The code `s1_burst_id.py` already have the correction implemented for this case [here](https://github.com/isce-framework/s1-reader/blob/ed16185338c879b59518ca26936751a1baffd621/src/s1reader/s1_burst_id.py#L90C9-L98C60), but the correction will be activated only if the `azimuthAnxTime` is longer than `T_orb`. This PR detects 2-3 ANX events before sensing time, calculates the crossing times, and pick the one closest one (but no later than) the sensing time. Therefore the ANX time is always less than `T_orb`.

To summarize in terms of the correction routine in `s1_burst_id.py`:
1. Making use of ANX time close to the sensing time was not able to exploit the correction logic implemented in `s1_burst_id.py`, so it resulted in incorrect calculations in both track numbers and ESA burst ID
2. The relative start / end relative track numbers are different. This causes the burst ID calculation to assume that ANX crossing has taken place in the data.
3. Incorrect track number was caused by choosing incorrect track number which is extracted from manifest. Having longer `azimuthAnxTime` than `T_orb` would have resulted in choosing the end relative orbit number, which is correct.
4. Incorrect ESA burst ID was caused because the calculation assumes ANX crossing in the data, (due to `2.`) it subtracts `T_orb`. This causes incorrect computation result for ESA burst ID.


Solution in plan:
- I think the logic in `s1_burst_id.py` is very robust, based on what we have seen when we were using annotation ANX time. This function was intensively tested when we’ve been using it for long time. Therefore I’m not so keen into changing the logics in the existing code.
- Therefore I would like to change the logic of choosing the orbit ANX time as being close to the annotation ANX time. Currently we are choosing the orbit ANX time closest to the sensing time (i.e. current implementation in the PR).

src/s1reader/s1_reader.py
seongsujeong: Make sense. I've changed the function `get_ascending_node_time_orbit` to raise `ValueError`. I'm replacing the snippet to use `try...except`, and make the code to use the annotation ANX time to make the workflow keep going.

src/s1reader/s1_reader.py
```diff
@@ -526,6 +529,83 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
...
+        # Convert the interpolated time into datetime
+        # Add the ascending node crossing time if it's before the sensing time
+        datetime_ascending_crossing = (datetime_orbit_ref
+                                        + datetime.timedelta(seconds=float(t_interp)))
+        if datetime_ascending_crossing < sensing_time:
+            datetime_ascending_node_crossing_list.append(datetime_ascending_crossing)
+
+    if len(datetime_ascending_node_crossing_list) == 0:
+        warnings.warn('Cannot detect ascending node crossings from the orbit information provided.')
+        return None
```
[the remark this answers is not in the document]
seongsujeong: Good suggestion. I've changed this function to raise `ValueError` rather than returning `None`.

src/s1reader/s1_reader.py
```diff
@@ -640,17 +720,21 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
         azimuth_time_interval = float(image_info_element.find('azimuthTimeInterval').text)
         slant_range_time = float(image_info_element.find('slantRangeTime').text)
-        ascending_node_time = as_datetime(image_info_element.find('ascendingNodeTime').text)
+        ascending_node_time_annotation =\
+            as_datetime(image_info_element.find('ascendingNodeTime').text)
+        first_line_utc_time = as_datetime(image_info_element.find('productFirstLineUtcTime').text)
 
-        downlink_element = tree.find('generalAnnotation/downlinkInformationList/downlinkInformation')
+        downlink_element = tree.find('generalAnnotation/downlinkInformationList/'
+                                     'downlinkInformation')
```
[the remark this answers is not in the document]
seongsujeong: I totally agree with your point about the formatting consistency, especially in the level of whole codes in this repo. Therefore I think another issue / PR would be necessary.

In case of this PR, I wanted to keep the lines in reasonable length suggested by PEP8.

seongsujeong:
@scottstanie The unit test for the ANX time computation is implemented into `test_orbit.py`. CircleCI as well as local pytest were successful.
Any suggestion about the code is appreciated. Thank you :)

src/s1reader/s1_reader.py
```diff
@@ -682,17 +785,48 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
...
+
+        except ValueError:
+            warnings.warn('Cannot estimate ascending node crossing time from Orbit. '
+                          'Using the ascending node time in annotation.')
+            ascending_node_time_orbit = None
+            ascending_node_time = ascending_node_time_annotation
+
+        # Calculate the difference in the two ANX times, and give
+        # warning message when the difference is noticeable.
+        if not ascending_node_time_orbit is None:
```
scottstanie, line 816:
```suggestion
        if ascending_node_time_orbit is not None:
```

https://peps.python.org/pep-0008/#programming-recommendations recommends to use `is not None`

src/s1reader/s1_reader.py
```diff
@@ -526,6 +529,100 @@ def get_track_burst_num(track_burst_num_file: str = esa_track_burst_id_file):
...
+
+    Returns
+    -------
+    _ : datetime.datetime
+        Ascending node crossing time calculated from orbit
+    '''
+
+    # Crop the orbit information before 2 * (orbit period) of sensing time
+    search_length = datetime.timedelta(seconds=2 * T_ORBIT)
+    orbit_until_sensing_time = get_burst_orbit(sensing_time  - search_length,
```
scottstanie, line 563: I think there's a way we could simplify this function by making `get_burst_orbit` a little more general (maybe make its hardcoded `pad_time` a function argument so you don't have to do the manual padding here). but not worth delaying this PR more

scottstanie (approved):
👍

seongsujeong:
@scottstanie Can you check if @gshiroma is a member of this organization `isce-framework`? I see his approval in this PR, but the check mask looks gray rather than green.

Issue #122, opened 2023-06-26
[Bug]: `BurstNoise` fails during `from_noise_annotation`


scottstanie:
I got this error
```python-traceback
...
  File "/u/aurora-r0/staniewi/repos/sweets/src/sweets/_geocode_slcs.py", line 47, in run_geocode
    cfg = GeoRunConfig.load_from_yaml(str(run_config_path), "s1_cslc_geo")
  File "/u/aurora-r0/staniewi/repos/COMPASS/src/compass/utils/geo_runconfig.py", line 96, in load_from_yaml
    bursts = runconfig_to_bursts(sns)
  File "/u/aurora-r0/staniewi/repos/COMPASS/src/compass/utils/runconfig.py", line 225, in runconfig_to_bursts
    for burst in load_bursts(safe_file, orbit_path, i_subswath, pol):
  File "/u/aurora-r0/staniewi/repos/s1-reader/src/s1reader/s1_reader.py", line 881, in load_bursts
    bursts = _burst_from_zip(path, id_str, orbit_path,
  File "/u/aurora-r0/staniewi/repos/s1-reader/src/s1reader/s1_reader.py", line 948, in _burst_from_zip
    bursts = burst_from_xml(f_annotation, orbit_path, f_tiff, iw2_f_annotation, z_file.open,
  File "/u/aurora-r0/staniewi/repos/s1-reader/src/s1reader/s1_reader.py", line 758, in burst_from_xml
    burst_noise = BurstNoise.from_noise_annotation(noise_annotation,
  File "/u/aurora-r0/staniewi/repos/s1-reader/src/s1reader/s1_annotation.py", line 840, in from_noise_annotation
    id_top = np.argmin(np.abs(noise_annotation.az_line-line_from))
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```

Note that  `id_top = np.argmin(np.abs(noise_annotation.az_line-line_from))` uses `.az_line`. This is not defined anywhere:
https://github.com/opera-adt/s1-reader/blob/ed16185338c879b59518ca26936751a1baffd621/src/s1reader/s1_annotation.py#L776-L789
it's probably a typo from copying this different class: https://github.com/opera-adt/s1-reader/blob/ed16185338c879b59518ca26936751a1baffd621/src/s1reader/s1_annotation.py#L274-L289

[...]

PR #123, opened 2023-06-28
Annotation class attrs

scottstanie:
This is trying to prevent #122 from popping up sporadically.
This is changing the `classmethod`s in the annotation module to return instances: https://stackoverflow.com/a/14605349/4174466
the functions that are decorated with `@classmethod` are really just alternate constructors. They're meant to create a new instance. Right now, they have been changing "class attributes", which are *shared across all instances of that class*.

*Why this is a problem*
If we kick of many instance of COMPASS in different threads, each one that reads bursts instantiates these annotation classes. they are all changing the same class attribute, so it's a data race where difference function calls to `.from_et` can clobber each other.

(example to show what this looks like)

<details>

```python
In [11]: from concurrent.futures import ThreadPoolExecutor
In [12]: import time, random

In [13]: class A:
    ...:     class_thing = 0

    ...:     @classmethod
    ...:     def from_x(cls, x):
    ...:         cls.class_thing = x

    ...:         # pretend we're doing some other parsing here in the construction
    ...:         time.sleep(random.random())

    ...:         # This is supposed to print the same thing twice:
    ...:         print(f"{x}, {cls.class_thing}")

    ...:         return cls()

In [14]: a1 = A.from_x(3)
3, 3

In [15]: a2 = A.from_x(2)
2, 2
```
if you try to do this method in many threads:
```python
In [24]: tpe = ThreadPoolExecutor(max_workers=20)

In [26]: alist = list(tpe.map(A.from_x, range(20)))
3, 19
18, 19
16, 19
7, 19
2, 19
8, 19
14, 19
6, 19
9, 19
12, 19
0, 19
11, 19
19, 19
5, 19
15, 19
4, 19
1, 19
10, 19
17, 19
13, 19
```

</details>

### DIFF
--- a/src/s1reader/s1_annotation.py
+++ b/src/s1reader/s1_annotation.py
@@ -96,12 +96,14 @@ class AnnotationBase:
     '''
     xml_et: ET
 
-    @classmethod
-    def _parse_scalar(cls, path_field: str, str_type: str):
+    @staticmethod
+    def _parse_scalar(xml_et: ET.ElementTree, path_field: str, str_type: str):
         '''A class method that parse the scalar value in AnnotationBase.xml_et
 
         Parameters
         ----------
+        xml_et : lxml.etree.ElementTree
+            XML tree to parse
         path_field : str
             Field in the xml_et to parse
         str_type : str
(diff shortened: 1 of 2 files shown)


seongsujeong (commented):
Thanks for the PR @scottstanie. I think the updates here are important to avoid the race condition when concurrent threads exists like the example you posted here.

Here are my comments and suggestions - Seongsu

src/s1reader/s1_annotation.py
```diff
@@ -419,56 +431,77 @@ def from_et(cls, et_in: ET):
...
+        number_of_samples = \
+            cls._parse_scalar(et_in,
+                              'imageAnnotation/imageInformation/numberOfSamples',
                               'scalar_int')
-        cls.number_of_samples = \
-            cls._parse_scalar('imageAnnotation/imageInformation/numberOfSamples',
+        number_of_samples = \
+            cls._parse_scalar(et_in,
+                              'imageAnnotation/imageInformation/numberOfSamples',
                               'scalar_int')
```
scottstanie, line 476:
```suggestion
```

PR #126, opened 2023-07-27
remove importlib_resources from `install_requires`

scottstanie:
Not sure why this was added, since [importlib.resources was available in python3.7](https://docs.python.org/3/library/importlib.resources.html) and we were never supporting older versions than that. We also dont seem to ever mention this package in the code.


LiangJYu (approved):
LGTM

seongsujeong (approved):
LGTM

PR #131, opened 2023-08-28
s1_info.py to take care of antimeridian-crossing

seongsujeong:
This PR is to take care of the case that the S1 frame's area crosses antimeridian when `s1_info.py` computes the bounding box. The fix is based on the code [here](https://github.com/nasa/opera-sds-pcm/pull/597/files)

Before:
>% python s1_info.py S1B_IW_SLC__1SDV_20211222T054456_20211222T054526_030134_039920_9391.zip --frame-bbox 
Found 1 Sentinel-1 SLC products.
S1B_IW_SLC__1SDV_20211222T054456_20211222T054526_030134_039920_9391.zip: [-177.710968, 50.39941, 179.223694, 52.575905]

After:
>% python s1_info.py S1B_IW_SLC__1SDV_20211222T054456_20211222T054526_030134_039920_9391.zip --frame-bbox 
Found 1 Sentinel-1 SLC products.
S1B_IW_SLC__1SDV_20211222T054456_20211222T054526_030134_039920_9391.zip: [178.642044, 50.39941, 182.728027, 52.575905]

### DIFF
--- a/src/s1reader/s1_info.py
+++ b/src/s1reader/s1_info.py
@@ -112,6 +112,11 @@ def _bounds_from_preview(safe_path: Union[Path, str]) -> list[float]:
     point_str = list(elem.text for elem in root.iter("coordinates"))[0]
     coords = [p.split(",") for p in point_str.split()]
     lons, lats = zip(*[(float(lon), float(lat)) for lon, lat in coords])
+
+    if max(lons) - min(lons) > 180.0:
+        # Antimeridian crossing detected
+        lons = [lon + (lon < 0) * 360.0 for lon in lons]
+
     return [min(lons), min(lats), max(lons), max(lats)]
 
 


gshiroma (approved):
LGTM. Thank you, @seongsujeong . I'm not familiar with that function. Please, make sure the any code that calls it can handle longitude values greater than +180 degrees.

seongsujeong:
AFAICT, neither `COMPASS` nor `RTC` is using this code in the workflow. `git grep` does not return anything from them. Therefore I think the PR should not affect the downstream workflows.

>(isce3_0.14.0) COMPASS% git grep _bounds_from_preview
(isce3_0.14.0) COMPASS% git grep s1_info             

(isce3_0.14.0) RTC% git grep _bounds_from_preview
(isce3_0.14.0) RTC% git grep s1_info

gshiroma:
> AFAICT, neither `COMPASS` nor `RTC` is using this code in the workflow. `git grep` does not return anything. Therefore I think the PR should not affect the downstream workflows.
> 
> > (isce3_0.14.0) COMPASS% git grep _bounds_from_preview
> > (isce3_0.14.0) COMPASS% git grep s1_info
> 
> > (isce3_0.14.0) RTC% git grep _bounds_from_preview
> > (isce3_0.14.0) RTC% git grep s1_info

Thank you, @seongsujeong . It looks like the function is used [here](https://github.com/isce-framework/s1-reader/blob/32875844a9c03dba4eba5bc31be2430d821f4453/src/s1reader/s1_info.py#L80), which seems to be used to inform the frame bbox [here](https://github.com/isce-framework/s1-reader/blob/32875844a9c03dba4eba5bc31be2430d821f4453/src/s1reader/s1_info.py#L255).

PR #132, opened 2023-08-29
Orbit file concatenation

scottstanie:
- started the function from what we discussed
- added the RESORBs for the edge case test
- did *not* account for the overlap between orbits as @seongsujeong pointed out. 

feel free to take over this one or start a new one


seongsujeong:
Thanks @scottstanie for the PR. I'll branch out from this, and add more features on top of that to take care of the edge case in RESORB.

Issue #42, opened 2022-05-20
[Bug]: `ET.parse(orbit_path)` FileNotFoundError


yunjunz:
### Checked for duplicates

Yes - I've already checked. Found by Ozzy Tirmizi.

### Describe the bug

```bash
>>> bursts = s1reader.load_bursts(zip_path, orbit_path, swath_num=2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/yunjunz/tools/s1-reader/src/s1reader/s1_reader.py", line 476, in load_bursts
    bursts = _burst_from_zip(path, id_str, orbit_path)
  File "/Users/yunjunz/tools/s1-reader/src/s1reader/s1_reader.py", line 536, in _burst_from_zip
    bursts = burst_from_xml(f_annotation, orbit_path, f_tiff, iw2_f_annotation, z_file.open)
  File "/Users/yunjunz/tools/s1-reader/src/s1reader/s1_reader.py", line 339, in burst_from_xml
    orbit_tree = ET.parse(orbit_path)
  File "/Users/yunjunz/tools/miniconda3/envs/opera/lib/python3.10/xml/etree/ElementTree.py", line 1229, in parse
    tree.parse(source, parser)
  File "/Users/yunjunz/tools/miniconda3/envs/opera/lib/python3.10/xml/etree/ElementTree.py", line 569, in parse
    source = open(source, "rb")
FileNotFoundError: [Errno 2] No such file or directory: ''
>>> 
```

### What did you expect?

I expected [...]

### Reproducible steps

```bash
import s1reader
zip_file = './SLC/S1A_IW_SLC__1SSV_20160829T230628_20160829T230655_012821_014394_A1F0.zip'
orbit_dir = './orbits'
orbit_path = s1reader.get_orbit_file_from_dir(zip_path, orbit_dir)
bursts = s1reader.load_bursts(zip_path, orbit_path, swath_num=2)
```

### Environment

```bash
- Version of this software [e.g. vX.Y.Z]
- Operating System: [e.g. MacOSX with Docker Desktop vX.Y]
...
```

yunjunz:
The issue I had is not finding the orbit file. After downloading it manually using isce2, the error is gone.

yunjunz:
We should print out a warning msg or raise error if `s1reader.get_orbit_file_from_dir()` returns empty.

hfattahi:
Yes I agree we should give a meaningful error message.

[...]

PR #133, opened 2023-08-30
Option to concatenate RESORB file

seongsujeong:
This PR, built upon @scottstanie's PR #132, attempts to solve the issue that can take place when using RESORB file with ascending node crossing (ANX) time correction.

RESORB file, which will be the major orbit information source for forward processing for OPERA, has shorter time range than POEORB file. It is about as long as two times of the orbit period (T_orb = 1:38:44.57), with time shifting of about T_orb. That means, In rare case, this can cause that no single RESORB file can cover the time range of `[sensing_start - T_orb, sensing_stop]`, which is necessary for ascending node crossing time correction. 
This PR attempts to find the two RESORB files, and concatenate them into one. The two source RESORB files are:
- One that covers `[sensing_start, sensing_end]` with padding (60 seconds)
- The other covers `[sensing_start - T_orb - padding, sensing_start - T_orb +padding]` to ensure ANX before sensing time.

The concatenation takes place either when trying to find the orbit files from directory (`get_orbit_file_from_dir()`), or when downloading the orbit file (`retrieve_orbit_file()`, which was `download_orbit()`)


src/s1reader/s1_orbit.py
```diff
@@ -341,3 +398,240 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+        Path to Sentinel1 SAFE zip file. Base names required to adhere to the
+        format described here:
+        https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/naming-conventions
+    orbit_file_list : list
+        List of the orbit files that exists in the system.
+
+    Returns:
+    --------
+    orbit_file : str
+        Path to the orbit file, or an empty string if no orbit file was found.
```
scottstanie, line 453:
```suggestion
        Path to the orbit file, or `None` if no orbit file was found.
```

src/s1reader/s1_orbit.py
```diff
@@ -341,3 +398,240 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    # string '.EOF' from end of end time string
+    t_orbit_stop = datetime.datetime.strptime(t_orbit_stop[:-4], FMT)
+
+    # check if:
+    # 1. swath start and stop time > orbit file start time
+    # 2. swath start and stop time < orbit file stop time
+    return all(t_orbit_start < t < t_orbit_stop for t in t_start_stop_frame)
+
+
+def concatenate_resorb_from_list(zip_path: str, orbit_file_list: list) -> str:
```
scottstanie, line 435:
```suggestion
def concatenate_resorb_from_list(zip_path: str, orbit_file_list: list) -> str | None:
```
It looks like we're returning either string or `None` rather than an empty string

src/s1reader/s1_orbit.py
```diff
@@ -56,16 +61,62 @@ def download_orbit(safe_file: str, orbit_dir: str):
...
+            #                 |---sensing time---|                          |---sensing time---|
+            #    CASE 1: adding earlier RESORB to latter            CASE 2: Adding latter RESORB to earlier
+            #                                                       (non-uniform temporal spacing takes place
+            #                                                      between `1` and `2` during the sensing time)
+
+            # adding earlier RESORB to latter (i.e. CASE 1 above)
+            concat_resorb_file = combine_xml_orbit_elements(orbit_file_indv_list[1],
+                                                            orbit_file_indv_list[0])
+
+    return concat_resorb_file
```
scottstanie, line 119: A couple questions
1. what does the `indv` mean in `orbit_file_indiv_list`? I can't immediately tell what that variable is going for
2. We have `if orbit_dict_list`, but that if statement is always `True`, whether or not `orbit_dict_earlier` or `orbit_dict_later` is None:

```python

In [1]: a = [None, None]

In [2]: if a:
   ...:     print("ok")
   ...:
ok
```
So maybe we should get rid of the `if`... but if we actually are testing something, we need an `else` because otherwise `concat_resorb_file` is not defined right now

src/s1reader/s1_orbit.py
```diff
@@ -131,8 +182,6 @@ def get_file_name_tokens(zip_path: str) -> [str, list[datetime.datetime]]:
     Returns
     -------
     mission_id: ('S1A', 'S1B')
-    orbit_path : str
```
scottstanie, line 134: (sorry for direct pushing to your branch, this docstring just confused me when i was looking through the code)

src/s1reader/s1_orbit.py
```diff
@@ -244,6 +293,9 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
         https://s1qc.asf.alaska.edu/aux_poeorb/
     auto_download : bool
         Automatically download the orbit file if not exist in the orbit_dir.
+    concat_resorb : bool
+        try to concatenate two RESORB files if there is no songle RESORB file that
```
scottstanie, line 297:
```suggestion
        try to concatenate two RESORB files if there is no single RESORB file that
```

src/s1reader/s1_orbit.py
```diff
@@ -267,6 +319,11 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
 
     orbit_file = get_orbit_file_from_list(zip_path, orbit_file_list)
 
+    # if no orbit file in the list, try to find RESORB files and concatenate
+    if not orbit_file and concat_resorb:
+        print('Attempting to concatenate RESORB files in the orbit file list.')
+        orbit_file = concatenate_resorb_from_list(zip_path, orbit_file_list)
+
     if orbit_file:
         return orbit_file
```
scottstanie, line 328: Just wanna check my understanding on the logic flow here:

1. we use `get_orbit_file_from_list` to check if an existing orbit in the directory covers the time span we need (a full orbit period + 60 seconds before the sensing start, sensing end)
2. then we try to concatenate two RESORBs from the list of existing ones with `concatenate_resorb_from_list`
3. then if that fails we call `retrieve_orbit_file`, which will also eventually concatenate them after downloading?

I think the part that threw me off at first was that we're checking for RESORB coverage in two different ways, once with the list of existing files and `_is_orbitfile_cover_timeframe`, and once with the `get_orbit_dict` checks.

src/s1reader/s1_orbit.py
```diff
@@ -56,16 +61,62 @@ def download_orbit(safe_file: str, orbit_dir: str):
...
+        if not os.path.exists(orbit_file):
+            download_orbit_file(orbit_dir, orbit_dict['orbit_url'])
 
-    return orbit_file
+        return orbit_file
+
+    # POEORB is not found, or there is no RESORB file that
+    # covers the sensing period + margin at the starting time.
+    # Try to find two subsequent RESORB files that covers the
+    # sensing period + margins at the startint time.
```
scottstanie, line 81:
```suggestion
    # sensing period + margins at the starting time.
```

scottstanie (commented):
Last question-

In `get_orbit_dict`, the arguments are ` start_time, end_time`, which each get padding added to them.
We're doing other specific padding to the start/end elsewhere in the code depending on the orbit type. Should we get rid of this? 
```python
    # Add a 30 min margin to start_time and end_time
    pad_30_min = datetime.timedelta(hours=0.5)
    pad_start_time = start_time - pad_30_min
    pad_end_time = end_time + pad_30_min
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    # 1. swath start and stop time > orbit file start time
+    # 2. swath start and stop time < orbit file stop time
+    return all(t_orbit_start < t < t_orbit_stop for t in t_start_stop_frame)
+
+
+def get_resorb_pair_from_list(zip_path: str, orbit_file_list: list,
+                               concatenate_resorb: bool=False) -> list | str | None:
+    '''
+    Find if there are TWO RESORB files that covers [start - margin_start_time, end]
+    If found, try to concatenate
```
gshiroma, line 450:
```suggestion
    If found, try to concatenate them
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    # check if:
+    # 1. swath start and stop time > orbit file start time
+    # 2. swath start and stop time < orbit file stop time
+    return all(t_orbit_start < t < t_orbit_stop for t in t_start_stop_frame)
+
+
+def get_resorb_pair_from_list(zip_path: str, orbit_file_list: list,
+                               concatenate_resorb: bool=False) -> list | str | None:
+    '''
+    Find if there are TWO RESORB files that covers [start - margin_start_time, end]
```
gshiroma, line 449:
```suggestion
    Find if there are two RESORB files that cover [start - margin_start_time, end]
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    If found, try to concatenate
+    Based on `get_orbit_file_from_list()`
+
+    Parameters:
+    -----------
+    zip_path : string
+        Path to Sentinel1 SAFE zip file. Base names required to adhere to the
```
gshiroma, line 456:
```suggestion
        Path to the Sentinel-1 SAFE zip file. Base names required to adhere to the
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+        format described here:
+        https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/naming-conventions
+    orbit_file_list : list
+        List of the orbit files that exists in the system.
```
gshiroma, line 460:
```suggestion
        List of the orbit files that exist in the system.
```
Maybe "List of the orbit files that exist locally in the system."?

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+
+    Returns:
+    --------
```
gshiroma, line 463:
```suggestion
    Returns
    -------
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+
+def get_resorb_pair_from_list(zip_path: str, orbit_file_list: list,
+                               concatenate_resorb: bool=False) -> list | str | None:
+    '''
+    Find if there are TWO RESORB files that covers [start - margin_start_time, end]
+    If found, try to concatenate
+    Based on `get_orbit_file_from_list()`
+
+    Parameters:
+    -----------
```
gshiroma, line 454:
```suggestion
    Parameters
    ----------
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+            # See NOTE in retrieve_orbit_file() for detail.
+            concat_resorb_filename = combine_xml_orbit_elements(resorb_filename_later,
+                                                                resorb_filename_earlier)
+            print('RESORB Concatenation successful.')
+            return concat_resorb_filename
+        else:
+            return [resorb_filename_earlier, resorb_filename_later]
+
+    print('Cannot find single RESORB file that meets the time frame criteria.')
+    return None
```
gshiroma, line 521:
```suggestion
    return
```
Not needed. It will "return" `None` regardless of you adding `None` or not

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    for resorb_file in resorb_file_list:
+        if mission_id not in os.path.basename(resorb_file):
+            continue
+
+        # 1. Try to find the orbit file that covers
+        # sensing time - margin_start_time
+        t_swath_start_stop_safe = [t_swath_start_stop[0] - pad_1min,
+                                   t_swath_start_stop[1] + pad_1min]
+        if _covers_timeframe(resorb_file, t_swath_start_stop_safe) and \
+            resorb_filename_later is None:
```
gshiroma, line 489:
```suggestion
        if (_covers_timeframe(resorb_file, t_swath_start_stop_safe) and
                resorb_filename_later is None):
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+            print('Found RESOEB file covering the S1 SAFE frame.')
+            resorb_filename_later = resorb_file
+            continue
+
+        # 2. Try to find the orbit file that covers the sensing start-stop
+        # with small padding (like 60 sec.)
+        t_swath_start_stop_anx = [t_swath_start_stop[0] - margin_start_time,
+                                  t_swath_start_stop[0] - margin_start_time + 2*pad_1min]
+        if _covers_timeframe(resorb_file, t_swath_start_stop_anx) and \
+            resorb_filename_earlier is None:
```
gshiroma, line 499:
```suggestion
        if (_covers_timeframe(resorb_file, t_swath_start_stop_anx) and
                resorb_filename_earlier is None):
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    Parameters
+    ----------
+    file1 : str
+        The path to the first .EOF file.
+    file2 : str
+        The path to the second .EOF file.
+
+    Returns
+    -------
+    str
```
gshiroma, line 540:
```suggestion
    _ : str
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+        The new final datetime of the updated orbital elements
+
+    Returns
+    -------
+    str
+        Generated filename.
+    """
+    product_name = Path(file_base).name
+    # >>> 'S1A_OPER_AUX_PREORB_OPOD_20200325T131800_V20200325T121452_20200325T184952'.index('V')
+    # 41
```
gshiroma, line 628:
```suggestion
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    fmt = "%Y%m%dT%H%M%S"
+    new_start_stop_str = new_start.strftime(fmt) + "_" + new_stop.strftime(fmt)
+    new_product_name = product_name[:42] + new_start_stop_str
+    return str(file_base).replace(product_name, new_product_name) + ".EOF"
+
+
+def _sort_list_of_osv(list_of_osvs):
+    '''
+    Sort the OSV with respect to the UTC time
```
gshiroma, line 637:
```suggestion
    Sort the orbit state vectors (OSV) with respect to the UTC time
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+
+    Parameters
+    ----------
+    list_of_osvs: ET.ElementTree
+        OSVs as XML ET
```
gshiroma, line 642:
```suggestion
        Orbit state vectors (OSVs) as XML ElementTree (ET) objects
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+
+    Returns
+    -------
+    list_of_osvs: ET.ElementTree
+        Sorted OSVs with respect to UTC
```
gshiroma, line 647:
```suggestion
        Sorted orbit state vectors (OSVs) with respect to UTC time
```

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,97 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
     return anx_time_orbit
 
 
+def get_osv_list_from_orbit(orbit_file: str | list,
+                            swath_start: datetime.datetime,
+                            swath_stop: datetime.datetime):
+    '''
+    Get the OSV list ET from `orbit file`
```
gshiroma, line 647:
```suggestion
    Get the list of orbit state vectors as ElementTree (ET) objects from the orbit file `orbit_file`
```

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,97 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
...
+
+    - when `orbit_file` is list, then all of the OSV lists are
```
gshiroma, line 649:
```suggestion
    - When `orbit_file` is a list, then all of the OSV lists are
```

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,97 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
...
+      get merged and sorted. The merging is done by concatenating
```
gshiroma, line 650:
```suggestion
      merged and sorted. The merging is done by concatenating
```

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,97 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
...
+      the OSVs to the "base OSV list".
+    - The "base OSV list" is the one that covers
+      the swath's start / stop time with padding applied,
+      so that the equal time spacing is guaranteed during that time perdiod.
```
gshiroma, line 654:
```suggestion
      so that the equal time spacing is guaranteed during that time period.
```

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,97 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
...
+
+    Parameters
+    ----------
+    orbit_file: str | list
+    orbit file's name, or list of the orbit file's nane
```
gshiroma, line 659:
```suggestion
        Orbit file name, or list of the orbit file names
```

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,97 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
...
+
+    swath_start: datetime.datetime
+        Sensing start time of the swath
+    swath_stop: datetime.datetime
+        Sensing start time of the swath
```
gshiroma, line 664:
```suggestion
        Sensing stop time of the swath
```

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,97 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
...
+    '''
+    if isinstance(orbit_file, str):
+        orbit_tree = ET.parse(orbit_file)
+        orbit_state_vector_list = orbit_tree.find('Data_Block/List_of_OSVs')
+        return orbit_state_vector_list
+
+    elif isinstance(orbit_file, list) and len(orbit_file):
+        # Concatenate the orbit files' OSV lists
+
+        #S1A_OPER_AUX_RESORB_OPOD_20230823T162050_V20230823T123139_20230823T154909.EOF
```
gshiroma, line 679:
```suggestion
```

src/s1reader/s1_reader.py
```diff
@@ -798,8 +903,11 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
 
     # find orbit state vectors in 'Data_Block/List_of_OSVs'
     if orbit_path:
-        orbit_tree = ET.parse(orbit_path)
-        orbit_state_vector_list = orbit_tree.find('Data_Block/List_of_OSVs')
+        #orbit_tree = ET.parse(orbit_path)
+        #orbit_state_vector_list = orbit_tree.find('Data_Block/List_of_OSVs')
```
gshiroma, line 907:
```suggestion
```

src/s1reader/s1_orbit.py
seongsujeong: 1. I meant 'individual` by `indv`, but I think we can just say like `orbit_file_indv_list` here.
2. Good point. Maybe I'll check `orbit_dict_list` with `all()`, and define the returning value with `else` statement.

tests/test_orbit.py
```diff
@@ -100,10 +105,44 @@ def test_anx_time(test_paths):
...
     diff_ascending_node_time_seconds = (ascending_node_time_orbit
                                         - ascending_node_time_annotation).total_seconds()
 
     assert abs(diff_ascending_node_time_seconds) < 0.5
+
+
+def test_combine_xml_orbit_elements(tmp_path, test_paths):
+    slc_file = tmp_path / "S1A_IW_SLC__1SDV_20230823T154908_20230823T154935_050004_060418_521B.SAFE"
+    slc_file.write_text("")
+    # test_paths.orbit_dir = f"{test_path}/data/orbits"
```
gshiroma, line 121:
```suggestion
```

tests/test_orbit.py
```diff
@@ -100,10 +105,44 @@ def test_anx_time(test_paths):
...
+    # The first attempt with only the normal RESORB files will fail,
+    # because there is no RESORB file in the directory
+    assert get_orbit_file_from_dir(slc_file, tmp_path, concat_resorb=False) is None
+
+    shutil.copy(orbit_dir / f1, tmp_path)
+    shutil.copy(orbit_dir / f2, tmp_path)
+
+    assert len(list(tmp_path.glob("*RESORB*.EOF"))) == 2
+
+    # When `concat_resorb` is False, then it returns the list of orbit files that
```
gshiroma, line 136:
```suggestion
    # When `concat_resorb` is `False`, then it returns the list of orbit files that
```

tests/test_orbit.py
```diff
@@ -100,10 +105,44 @@ def test_anx_time(test_paths):
...
+    # covers the sensing time + last ANX time before sensing start
+    resorb_file_list = get_orbit_file_from_dir(slc_file, tmp_path, concat_resorb=False)
+    resorb_file_basename_list = [os.path.basename(filename) for filename in resorb_file_list]
+    resorb_file_basename_list.sort()
+    assert resorb_file_basename_list == [f1, f2]
+
+    # When `concat_resorb` is True, when it combines two orbit file into two
```
gshiroma, line 143:
```suggestion
    # When `concat_resorb` is `True`, when it combines two orbit file into two
```

gshiroma (commented):
Thank you for working on that @seongsujeong . I have some comments for now.

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    -------
+    str
+        Generated filename.
+    """
+    product_name = Path(file_base).name
+    # >>> 'S1A_OPER_AUX_PREORB_OPOD_20200325T131800_V20200325T121452_20200325T184952'.index('V')
+    # 41
+    fmt = "%Y%m%dT%H%M%S"
+    new_start_stop_str = new_start.strftime(fmt) + "_" + new_stop.strftime(fmt)
+    new_product_name = product_name[:42] + new_start_stop_str
```
gshiroma, line 631: The hard-coded number `42` seems dangerous to me. A way to make it a bit more robust would be to split the `product_name` using underscores `_`.

src/s1reader/s1_orbit.py
seongsujeong: There has been a change in the logic flow. based on the SAS tests and the discussions from ADT/SDS meeting
- We are going to provide the RESORB pair as a list. I've kept the orbit file concatenation feature, but it will be turned off by default.
- search for RESORB pair will be attempted in `concatenate_resorb_from_list` and `retrieve_orbit_file` because some R2 SAS uses `concatenate_resorb_from_list()`. Doing the RESORB pair search in `get_orbit_file_from_dir` level will make the SAS not be able to find the RESORB pair.
- In searching for existing orbit file, `get_resorb_pair_from_list` will be used. `get_orbit_dict` will be used for finding the orbit files from internet.

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +399,267 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    '''
+    Check if `orbitfile` covers `t_start_stop_frame`
+    Copied from `get_orbit_file_from_list()` and modified
+
+    Parameter
+    ---------
+    orbit_file: str
+
+    t_start_stop_frame: list(datetime.datetime)
+
```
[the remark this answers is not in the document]
seongsujeong: Populated the parameter's description accordingly.

src/s1reader/s1_orbit.py
seongsujeong: Docstring was revised accordingly. Thank you.

src/s1reader/s1_orbit.py
seongsujeong: Docstring was revised accordingly. Thank you.

src/s1reader/s1_orbit.py
seongsujeong: Removing the description about how `41` was determined, but I think it would be a good reference of keep the example orbit filename here.

src/s1reader/s1_orbit.py
seongsujeong: Agreed. I've changed the logic here to find a index for `_V`, and make use of that to slice the string `product_name`

seongsujeong:
Thank you @scottstanie and @gshiroma for taking a look at the PR and making a good points in the comment. Would you mind taking another look? Thanks.

seongsujeong:
> Last question-
> 
> In `get_orbit_dict`, the arguments are ` start_time, end_time`, which each get padding added to them. We're doing other specific padding to the start/end elsewhere in the code depending on the orbit type. Should we get rid of this?
> 
> ```python
>     # Add a 30 min margin to start_time and end_time
>     pad_30_min = datetime.timedelta(hours=0.5)
>     pad_start_time = start_time - pad_30_min
>     pad_end_time = end_time + pad_30_min
> ```

based on my cartoon-based thought, 2 * 30 mins of padding would be okay for timing design of RESORB (duration, time shifting). However, shorter padding might help reducing the chance of the unnecessary use of RESORB pair. Based on the personal conversation with @vbrancat and time constrain we have, I prefer not to touch this part.

scottstanie:
Sure not changing it now due to the time crunch is fine. I just wanted to make sure we know that we are padding the start once with `T_ORBIT + 60 seconds` and then again with 30 minutes in a different function

src/s1reader/s1_orbit.py
gshiroma: Sounds good. Thank you, @seongsujeong . Maybe you can remove now `.index('V')` and `#42`

src/s1reader/s1_orbit.py
```diff
@@ -56,16 +61,69 @@ def download_orbit(safe_file: str, orbit_dir: str):
...
+            # adding earlier RESORB to latter (i.e. CASE 1 above)
+            if concatenate:
+                concat_resorb_file = combine_xml_orbit_elements(resorb_file_list[1],
+                                                                resorb_file_list[0])
+                return concat_resorb_file
+
+            return resorb_file_list
+
+        warnings.warn('Cannot find RESORB pair that covers the timeframe in the query')
+        return None
```
gshiroma, line 126:
```suggestion
        return
```

src/s1reader/s1_orbit.py
```diff
@@ -56,16 +61,69 @@ def download_orbit(safe_file: str, orbit_dir: str):
...
+            download_orbit_file(orbit_dir, orbit_dict['orbit_url'])
 
-    return orbit_file
+        return orbit_file
+
+    # POEORB is not found, or there is no RESORB file that
+    # covers the sensing period + margin at the starting time.
+    # Try to find two subsequent RESORB files that covers the
+    # sensing period + margins at the starting time.
+    if orbit_dict is None:
```
gshiroma, line 82: This if-statement here looks useless to me. I think at this point `orbit_dict` cannot be `not None`. Otherwise, it would have entered the if-statement in line `71` and escaped in line `76`.

src/s1reader/s1_orbit.py
```diff
@@ -56,16 +61,69 @@ def download_orbit(safe_file: str, orbit_dir: str):
...
+                                            start_time + 2 * pad_short, 'AUX_RESORB')
+
+        resorb_dict_later = get_orbit_dict(mission_id,
+                                          start_time + margin_start_time - pad_short,
+                                          end_time + pad_short,
+                                          'AUX_RESORB')
+
+        resorb_dict_list = [resorb_dict_earlier, resorb_dict_later]
+
+        if all(resorb_dict_list):
```
gshiroma, line 96: This will simplify your code:

```suggestion
        if not all(resorb_dict_list):
            warnings.warn('Cannot find RESORB pair that covers the timeframe in the query')
            return
```

And indent the rest of the block to the left.

src/s1reader/s1_orbit.py
```diff
@@ -244,6 +302,9 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
         https://s1qc.asf.alaska.edu/aux_poeorb/
     auto_download : bool
         Automatically download the orbit file if not exist in the orbit_dir.
+    concat_resorb : bool
+        try to concatenate two RESORB files if there is no single RESORB file that
```
gshiroma, line 306:
```suggestion
        Concatenate two RESORB files if there is no single RESORB file that
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +398,271 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    # check if:
+    # 1. swath start and stop time > orbit file start time
+    # 2. swath start and stop time < orbit file stop time
+    return all(t_orbit_start < t < t_orbit_stop for t in t_start_stop_frame)
+
+
+def get_resorb_pair_from_list(zip_path: str, orbit_file_list: list,
+                               concatenate_resorb: bool=False) -> list | str | None:
+    '''
+    Find if there are TWO RESORB files that cover [start - margin_start_time, end]
```
gshiroma, line 450:
```suggestion
    Find if there are two RESORB files that cover [start - margin_start_time, end]
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +398,271 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+            # BE CAREFUL ABOUT THE ORDER HOW THEY ARE CONCATENATED.
+            # See NOTE in retrieve_orbit_file() for detail.
+            concat_resorb_filename = combine_xml_orbit_elements(resorb_filename_later,
+                                                                resorb_filename_earlier)
+            print('RESORB Concatenation successful.')
+            return concat_resorb_filename
+        else:
+            return [resorb_filename_earlier, resorb_filename_later]
+
+    print('Cannot find single RESORB file that meets the time frame criteria.')
```
gshiroma, line 522:
```suggestion
    print('Cannot find a single RESORB file that meets the time frame criteria.')
```

gshiroma (commented):
A few more comments. Tomorrow, I'll review the logic more in-depth.

src/s1reader/s1_orbit.py
seongsujeong: makes sense. I've removed the it statement here.

src/s1reader/s1_orbit.py
seongsujeong: good point. we can do early return here.

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +397,271 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+        if (_covers_timeframe(resorb_file, t_swath_start_stop_safe) and
+                resorb_filename_later is None):
+            print('Found RESOEB file covering the S1 SAFE frame.')
+            resorb_filename_later = resorb_file
+            continue
+
+        # 2. Try to find the orbit file that covers the sensing start-stop
+        # with small padding (like 60 sec.)
+        t_swath_start_stop_anx = [t_swath_start_stop[0] - margin_start_time,
+                                  t_swath_start_stop[0] - margin_start_time + 2*pad_1min]
```
hfattahi, line 498: I expected to search for time range of "SLC_sensing_start - T_ORB" to "SLC_sensing_start - pad_1min". 
But instead we are searching an orbit to cover "SLC_sensing_start - T_ORB" to "SLC_sensing_start - T_ORB + 2 pad_1min". Again was this the intention really?

hfattahi (commented):
Sending my partial review.

seongsujeong:
TODO: get rid of 30 min padding in `get_orbit_dict()`

src/s1reader/s1_orbit.py
```diff
@@ -277,24 +337,24 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
         return
 
     # Attempt auto download
-    orbit_file = download_orbit(zip_path, orbit_dir)
+    orbit_file = retrieve_orbit_file(zip_path, orbit_dir, concat_resorb)
     return orbit_file
 
 
-def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
+def get_orbit_file_from_list(zip_path: str, orbit_file_list: list, concat_resorb: bool=False) -> str | list | None:
```
scottstanie, line 344: Note about this: Sorting `orbit_file_list` would be useful for reproducibility

src/s1reader/s1_orbit.py
```diff
@@ -277,24 +337,24 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
...
     orbit_file_list : list
         List of the orbit files that exists in the system.
 
-    Returns:
-    --------
+    Returns
+    -------
     orbit_file : str
         Path to the orbit file, or an empty string if no orbit file was found.
     '''
```
scottstanie, line 360:
```
t_swath_start_stop[0] = t_swath_start_stop[0] - margin_start_time
```
similar to heresh's naming comming, maybe we can unpack this pair of times and name them more specifically

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,96 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
...
+    -------
+    orbit_state_vector_list: ET
+        Orbit state vector list
+    '''
+    if isinstance(orbit_file, str):
+        orbit_tree = ET.parse(orbit_file)
+        orbit_state_vector_list = orbit_tree.find('Data_Block/List_of_OSVs')
+        return orbit_state_vector_list
+
+    elif isinstance(orbit_file, list) and len(orbit_file):
```
scottstanie, line 676:
```suggestion
    elif isinstance(orbit_file, list) and len(orbit_file) == 1:
        # ... do the simple reading like above
    elif isinstance(orbit_file, list) and len(orbit_file) > 1:
```

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,96 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
...
+
+        for filename in orbit_file:
+            start_time_str = \
+                os.path.basename(filename).split('_')[6].replace('V','')
+            orbit_start_time_list.append(datetime.datetime.strptime(start_time_str,
+                                                                    orbitfile_datetime_format))
+            stop_time_str = \
+                os.path.basename(filename).split('_')[7].replace('V','').replace('.EOF','')
+            orbit_stop_time_list.append(datetime.datetime.strptime(stop_time_str,
+                                                                   orbitfile_datetime_format))
```
scottstanie, line 693: we should have some function that gets the start and/or stop time elsewhere in the code because it's so common

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,96 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
...
+                          f'{PADDING_TIME_ORBIT} seconds of padding. '
+                          'Using the latest orbit as the base.')
+
+            sorted_index_list = [index for index, _ in
+                                sorted(enumerate(orbit_start_time_list), key=lambda x: x[1])]
+
+            orbit_list_ordered = [orbit_file[index] for index in sorted_index_list]
+
+            # Load the last orbit file's OSV list
+            base_orbit_file_index = orbit_list_ordered[-1]
```
scottstanie, line 713: this might be wrong

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,96 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
     return anx_time_orbit
 
 
+def get_osv_list_from_orbit(orbit_file: str | list,
```
scottstanie, line 643: This function should just read in the orbits from the `orbit_file` or the list of orbit files

scottstanie: Let's assume that if we get a list, it is sorted by start time

src/s1reader/s1_orbit.py
```diff
@@ -15,20 +16,24 @@
...
-margin_start_time = datetime.timedelta(seconds=T_ORBIT + 60.0)
+margin_start_time = datetime.timedelta(seconds=T_ORBIT + PADDING_SHORT)
 
 # Scihub guest credential
 scihub_user = 'gnssguest'
 scihub_password = 'gnssguest'
 
 
-def download_orbit(safe_file: str, orbit_dir: str):
+def retrieve_orbit_file(safe_file: str, orbit_dir: str, concatenate: bool=False):
```
scottstanie, line 31: let's make sure this returns a file list that's sorted by start time

src/s1reader/s1_orbit.py
seongsujeong: I would like to keep the original code that represents the time window as a list. So I'm making the list's name more clear that it's about the search window, not the sensing start / stop.

```python
if all([t_orbit_start < t < t_orbit_stop for t in t_search_window]):
```

src/s1reader/s1_orbit.py
```diff
@@ -337,7 +397,271 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    resorb_filename_earlier = None
+    resorb_filename_later = None
+    for resorb_file in resorb_file_list:
+        if mission_id not in os.path.basename(resorb_file):
+            continue
+
+        # 1. Try to find the orbit file that covers
+        # sensing time - margin_start_time
+        t_swath_start_stop_safe = [t_swath_start_stop[0] - pad_1min,
+                                   t_swath_start_stop[1] + pad_1min]
```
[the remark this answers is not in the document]
seongsujeong: I think the confusion here is from the incorrect comment. They are swapped to each other. I've fixed the comment around here.

src/s1reader/s1_reader.py
seongsujeong: Considering this as resolved as I've removed the code above.

src/s1reader/s1_orbit.py
seongsujeong: It will return the list of the orbit files in ascending order:
```python
    resorb_dict_list = [resorb_dict_earlier, resorb_dict_later]

    if not all(resorb_dict_list):
        warnings.warn('Cannot find RESORB pair that covers the timeframe in the query')
        return

    # Either concatenate the orbit, or return the RESORB pair
    resorb_file_list = []
    for resorb_dict in resorb_dict_list:
        resorb_file = os.path.join(orbit_dir, f"{resorb_dict['orbit_name']}.EOF")
        .
        .
        .
        resorb_file_list.append(resorb_file)
    .
    .
    .
    return resorb_file_list
```

src/s1reader/s1_orbit.py
```diff
@@ -56,16 +61,68 @@ def download_orbit(safe_file: str, orbit_dir: str):
     # Find precise orbit first
```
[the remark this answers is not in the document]
seongsujeong: Good point. I've differentiated variable names here by using prefix `sensing_` or `search_`

seongsujeong:
@scottstanie @hfattahi @gshiroma Thank you for the good comments and feedback. I think this PR is ready for the another round of review.

src/s1reader/s1_reader.py
```diff
@@ -19,17 +19,21 @@
...
 
 from s1reader.s1_burst_slc import Doppler, Sentinel1BurstSlc
 from s1reader.s1_burst_id import S1BurstId
-from s1reader.s1_orbit import T_ORBIT
+from s1reader.s1_orbit import (T_ORBIT, merge_osv_list)
+from s1reader.s1_orbit import FMT as orbitfile_datetime_format
 
 # Tolerance of the ascending node crossing time in seconds
 ASCENDING_NODE_TIME_TOLERANCE_IN_SEC = 1.0
+PADDING_TIME_ORBIT = 60.0
```
LiangJYu, line 36: Is this padding meant to be the same as `PADDING_SHORT` in `s1_orbit.py`? If so, let's move it there so there's a single parameter instead two that do the same.

src/s1reader/s1_orbit.py
```diff
@@ -332,12 +398,278 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    merge_osv_list(list_of_osvs1, list_of_osvs2)
+
+    outfile = _generate_filename(file1, new_start_dt, new_stop_dt)
+    tree1.write(outfile, encoding="UTF-8", xml_declaration=True)
+    return outfile
+
+
+def merge_osv_list(list_of_osvs1, list_of_osvs2):
+    # Extract the UTC from the OSV of the first XML
+    osv1_utc_list = [datetime.datetime.fromisoformat(osv1.find('UTC').text.replace('UTC=',''))
```
LiangJYu, line 622: Let's make `datetime.datetime.fromisoformat(osv1.find('UTC').text.replace('UTC=','')` it's own function in `s1_orbit.py` because it's repeated several times in this PR. Maybe call it `strip_UTC_from_time_string`?

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,67 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
...
+      the OSVs to the "base OSV list".
+    - The "base OSV list" is the one that covers
+      the swath's start / stop time with padding applied,
+      so that the equal time spacing is guaranteed during that time period.
+
+    Parameters
+    ----------
+    orbit_file: str | list
+        Orbit file name, or list of the orbit file names
+
```
LiangJYu, line 660:
```suggestion
```
formatting nit

src/s1reader/s1_reader.py
```diff
@@ -627,6 +640,67 @@ def get_ascending_node_time_orbit(orbit_state_vector_list: ET,
...
+            src_orbit_tree = ET.parse(src_orbit_filename)
+            src_osv_vector_list = src_orbit_tree.find('Data_Block/List_of_OSVs')
+            orbit_state_vector_list = merge_osv_list(orbit_state_vector_list,
+                                                     src_osv_vector_list)
+
+        return orbit_state_vector_list
+
+    else:
+        raise RuntimeError('Invalid orbit information provided. '
+                           'It has to be filename or the list of filename.')
```
LiangJYu, line 702:
```suggestion
                           'It has to be filename or the list of filenames.')
```
nit

tests/test_orbit.py
```diff
@@ -3,14 +3,19 @@
...
 import datetime
+import os
+from pathlib import Path
+import shutil
 import zipfile
 
 import isce3
 import lxml.etree as ET
 import numpy as np
+import pytest
```
LiangJYu, line 14: Appears to be unused
```suggestion
```

LiangJYu (commented):
Mostly nits. Sorry, some of my comments may stale.

hfattahi (approved):
LGTM. Thanks @seongsujeong .

src/s1reader/s1_reader.py
seongsujeong: `s1_orbit.py` has a constant `PADDING_SHORT` that does the similar thing. I've changed this code to import the constant from `s1_orbit.py`

src/s1reader/s1_orbit.py
```diff
@@ -274,27 +336,27 @@ def get_orbit_file_from_dir(zip_path: str, orbit_dir: str, auto_download: bool =
...
     zip_path : string
         Path to Sentinel1 SAFE zip file. Base names required to adhere to the
         format described here:
         https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/naming-conventions
     orbit_file_list : list
         List of the orbit files that exists in the system.
 
-    Returns:
-    --------
+    Returns
```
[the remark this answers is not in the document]
seongsujeong: Good point. I've added the parameter description for other functions as well.

src/s1reader/s1_orbit.py
seongsujeong: Good idea. I've defined a function `_get_uct_time_from_osv()` that extracts UTC time from osv.

src/s1reader/s1_orbit.py
```diff
@@ -6,6 +6,7 @@
 import os
 import requests
 import warnings
+from pathlib import Path
```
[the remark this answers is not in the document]
seongsujeong: Adjusted the module import order based on the results from `pylint`

src/s1reader/s1_orbit.py
```diff
@@ -56,16 +61,65 @@ def download_orbit(safe_file: str, orbit_dir: str):
...
+        orbit_dict_later = get_orbit_dict(mission_id,
+                                          start_time + margin_start_time - pad_short,
+                                          end_time + pad_short,
+                                          'AUX_RESORB')
+
+        orbit_dict_list = [orbit_dict_earlier, orbit_dict_later]
+
+        if orbit_dict_list:
+            orbit_file_indv_list = []
+            for orbit_dict in orbit_dict_list:
```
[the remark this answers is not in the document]
seongsujeong: In case that we don't find the RESORB pair, the function will send out the warning message and return `None`. It seems like that the overall behavior of this module is to return `None` when there is no orbit file is found, so wanted to let this new feature to behave the same.

src/s1reader/s1_orbit.py
```diff
@@ -56,16 +61,65 @@ def download_orbit(safe_file: str, orbit_dir: str):
...
+            #                                            (non-uniform temporal spacing takes place
+            #                                          between `1` and `2` during the sensing time)
+            #
+            # adding earlier RESORB to latter (i.e. CASE 1 above)
+            if concatenate:
+                concat_resorb_file = combine_xml_orbit_elements(orbit_file_indv_list[1],
+                                                                orbit_file_indv_list[0])
+                return concat_resorb_file
+
+            return orbit_file_indv_list
```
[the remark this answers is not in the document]
seongsujeong: In case that we don't find the RESORB pair, the function will send out the warning message and return `None`. The snippet is like below. It seems like that the overall behavior of this module is to return `None` when there is no orbit file is found, so wanted to let this new feature to behave the same.

```python
    if not all(resorb_dict_list):
        warnings.warn('Cannot find RESORB pair that covers the timeframe in the query')
        return
```

src/s1reader/s1_orbit.py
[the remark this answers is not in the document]
seongsujeong: Right above this if statement, we check if all the elements in `orbit_dic_list` (which is `resorb_dict_list` now) are valid.  If not, we early return. The snippet for that is as below. 

```python
    if not all(resorb_dict_list):
        warnings.warn('Cannot find RESORB pair that covers the timeframe in the query')
        return
```

Therefore all the elements in the list has valid values at this point.

seongsujeong:
@LiangJYu Thanks for the feedback. All good points. Please let me know if you have more suggestions, or if I have missed responding some of your comments. Thank you!

src/s1reader/s1_orbit.py
```diff
@@ -332,12 +417,323 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    merge_osv_list(list_of_osvs1, list_of_osvs2)
+
+    outfile = _generate_filename(file1, new_start_dt, new_stop_dt)
+    tree1.write(outfile, encoding="UTF-8", xml_declaration=True)
+    return outfile
+
+
+def merge_osv_list(list_of_osvs1, list_of_osvs2):
+    '''
+    Merge the two orvit state vector list (OSV list) in orbit files into one.
```
scottstanie, line 629:
```suggestion
    Merge the two orbit state vector list (OSV list) in orbit files into one.
```

src/s1reader/s1_orbit.py
```diff
@@ -170,10 +241,8 @@ def get_orbit_dict(mission_id, start_time, end_time, orbit_type):
...
         raise ValueError(err_msg)
 
-    # Add a 30 min margin to start_time and end_time
-    pad_30_min = datetime.timedelta(hours=0.5)
-    pad_start_time = start_time - pad_30_min
-    pad_end_time = end_time + pad_30_min
+    pad_start_time = start_time
+    pad_end_time = end_time
     new_start_time = pad_start_time.strftime('%Y-%m-%dT%H:%M:%S')
     new_end_time = pad_end_time.strftime('%Y-%m-%dT%H:%M:%S')
```
scottstanie, line 247:
```suggestion
    new_start_time = start_time.strftime('%Y-%m-%dT%H:%M:%S')
    new_end_time = end_time.strftime('%Y-%m-%dT%H:%M:%S')
```

src/s1reader/s1_orbit.py
```diff
@@ -170,10 +241,8 @@ def get_orbit_dict(mission_id, start_time, end_time, orbit_type):
         err_msg = f'{orbit_type} not a valid orbit type'
         raise ValueError(err_msg)
```
scottstanie, line 243: Can't comment on the docstring but maybe we should change it now from 

```python
    start_time: datetime object
        Sentinel start acquisition time
    end_time: datetime object
        Sentinel end acquisition time
```

to

```python
    start_time: datetime object
        Starting time for search query
    end_time: datetime object
        Ending time  for search query
```

since this function really doesn't care about sensing times; it just makes the query for whatever you give (which in our case is not the start/stop senting time

src/s1reader/s1_orbit.py
```diff
@@ -332,12 +417,323 @@ def get_orbit_file_from_list(zip_path: str, orbit_file_list: list) -> str:
...
+    return
+
+
+def combine_xml_orbit_elements(file1: str, file2: str) -> str:
+    """Combine the orbit elements from two XML files.
+
+    Create a new .EOF file with the combined results.
+    Output is named with the start_datetime and stop_datetime changed, with
+    the same base as `file1`.
+
```
scottstanie, line 573:
```suggestion

    `file1` is the "base" of the output: All of the orbit state vectors from `file1` are used, 
    and only the OSVs from `file2` which do fall within the time range of `file1` are included.

    Create a new .EOF file with the combined results.
    Output is named with the start_datetime and stop_datetime changed..

```
clarifying what "same base as file1" means

scottstanie (approved):
good cutting down on the possible code paths for this.
general note for a later refactor- we should make a give of the functions in `s1_orbit`  (and moreso in `s1_reader`) leading underscore names to indicate that they aren't public

src/s1reader/s1_orbit.py
seongsujeong: Good catch. The docstring needs to be revised.

seongsujeong:
Thank you @scottstanie, @gshiroma, @hfattahi, and @LiangJYu for the review! I think the code is much better than the initial commit.

Issue #118, opened 2023-05-27
Bad number of boundary/center points extracted


scottstanie:
Steps to reproduce

```
$ asfsmd S1A_IW_SLC__1SDV_20150124T165546_20150124T165613_004316_005423_ADFE.zip
$ s1_info S1A_IW_SLC__1SDV_20150124T165546_20150124T165613_004316_005423_ADFE.SAFE/
```

Error:
```python-traceback
Traceback (most recent call last):
  File "/Users/staniewi/miniconda3/envs/mapping/bin/s1_info", line 8, in <module>
    sys.exit(main())
  File "/Users/staniewi/repos/s1-reader/src/s1reader/s1_info.py", line 262, in main
    for burst in get_bursts(path, args.pol, args.iw):
  File "/Users/staniewi/repos/s1-reader/src/s1reader/s1_info.py", line 26, in get_bursts
    burst_nested_list = [
  File "/Users/staniewi/repos/s1-reader/src/s1reader/s1_info.py", line 27, in <listcomp>
    s1reader.load_bursts(filename, None, iw, pol, flag_apply_eap=False)
  File "/Users/staniewi/repos/s1-reader/src/s1reader/s1_reader.py", line 878, in load_bursts
    bursts = _burst_from_safe_dir(path, id_str, orbit_path,
  File "/Users/staniewi/repos/s1-reader/src/s1reader/s1_reader.py", line 997, in _burst_from_safe_dir
    bursts = burst_from_xml(f_annotation, orbit_path, f_tiff, iw2_f_annotation,
  File "/Users/staniewi/repos/s1-reader/src/s1reader/s1_reader.py", line 795, in burst_from_xml
    platform_id, safe_filename, center_pts[i],
IndexError: list index out of range
```

Inside `get_burst_centers_and_boundaries`:

```
ipdb> p center_pts
[<POINT (14.688 36.912)>, <POINT (14.649 37.078)>, <POINT (14.61 37.244)>, <POINT (14.571 37.41)>, <POINT (14.531 37.575)>, <POINT (14.492 37.741)>, <POINT (14.453 37.907)>, <POINT (14.409 38.09)>]
ipdb> p len(center_pts)
8
```

but,
```
ipdb> p len(burst_list_elements)
9
```

scottstanie:
more failing zips over italy, all early in 2015

```
S1A_IW_SLC__1SDV_20150205T165551_20150205T165618_004491_00581E_9231.zip
[...]

PR #134, opened 2023-09-12
Patch for "no polynomial" case for azimuth FM rate mitigation & handling the burst without burst polygon

seongsujeong:
This PR is an attempt to fix the rare case that the list pf polynomials (azimuth FM rate or doppler centroid) does not have any overlap with the sensing start / stop of the burst. In that case, the azimuth FM rate mismatch correction finds the nearest polynomials in azimuth time, but the indices for polynomials (`index_start` and `index_end`) has the same value. When that happens interpolator from `InterpolatedUnivariateSpline` cannot be generated.

When that happens, the `index_start` and `index_end` are adjusted so that an`InterpolatedUnivariateSpline` gets to have valid parameters to create the interpolator.
When using the generated interpolator, optional parameter `ext=3` was set so that the interpolator returns the values at the left / right boundary values when the input parameter for the interpolation of out of range. This is equivalent to nearest neighbor in this case because the sensing start / stop is "out of range" from the list of polynomials' azimuth time.

Another issue that this PR handles is the when the burst does not have corresponding burst polygon. Such cases have been discovered as documented in #118. When that is the case, s1-reader will fill `Sentinel1BurstSlc.border` member with empty polygon. Likewise `Sentinel1BurstSlc.center` will be populated by an empty point geometry.

### DIFF
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -912,6 +912,9 @@ class Sentinel1BurstSlc:
 
         # calculate splined interpolation of the coeffs. and tau_0s
         if len(fm_rate_aztime_sec_vec) <= 1:
+            # Interpolator object cannot be created with only one set of polynomials.
+            # Such case happens when there is no polygon that falls between a burst's start / stop
+            # Return an empty LUT2d in that case
             return isce3.core.LUT2d()
         interpolator_tau0_ka = InterpolatedUnivariateSpline(
                                         fm_rate_aztime_sec_vec,
@@ -921,6 +924,9 @@ class Sentinel1BurstSlc:
         tau0_ka_interp = interpolator_tau0_ka(vec_t)[..., np.newaxis]
 
         if len(dc_aztime_sec_vec) <=1:
+            # Interpolator object cannot be created with only one set of polynomials.
+            # Such case happens when there is no polygon that falls between a burst's start / stop
+            # Return an empty LUT2d in that case
             return isce3.core.LUT2d()
         interpolator_tau0_fdc_interp = InterpolatedUnivariateSpline(
                                         dc_aztime_sec_vec,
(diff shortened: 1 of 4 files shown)


seongsujeong:
@vbrancat @LiangJYu I think the PR is ready to review. I will submit a PR to COMPASS To change the code accordingly.

src/s1reader/s1_burst_slc.py
```diff
@@ -911,15 +911,21 @@ def az_fm_rate_mismatch_from_llh(self,
...
+            return isce3.core.LUT2d()
         interpolator_tau0_ka = InterpolatedUnivariateSpline(
                                         fm_rate_aztime_sec_vec,
                                         self.extended_coeffs.fm_rate_tau0_vec,
+                                        ext=3,
                                         k=1)
         tau0_ka_interp = interpolator_tau0_ka(vec_t)[..., np.newaxis]
 
+        if len(dc_aztime_sec_vec) <=1:
+            return isce3.core.LUT2d()
```
seongsujeong, line 924: add more description about why empty LUT2 is being returned
Also some description about why such treatment was necesary(i.e. historical data w/ IPF 002.36)

src/s1reader/s1_reader.py
```diff
@@ -14,6 +14,7 @@
 import isce3
 import numpy as np
 import shapely
+from shapely.affinity import translate
```
seongsujeong, line 17: get rid of this module

vbrancat (approved):
LGTM @seongsujeong. We reviewed this PR during our OPERA CSLC tag-up meeting. Please, add comments in the code on which data is affected by this fix. Thank you :)

src/s1reader/s1_burst_slc.py
```diff
@@ -911,15 +911,21 @@ def az_fm_rate_mismatch_from_llh(self,
                              for datetime_vec in self.extended_coeffs.dc_aztime_vec]
 
         # calculate splined interpolation of the coeffs. and tau_0s
+        if len(fm_rate_aztime_sec_vec) <=1:
```
LiangJYu, line 914:
```suggestion
        if len(fm_rate_aztime_sec_vec) <= 1:
```
nit - formatting. Same below

LiangJYu (approved):
LGTM. Only a nit comment

PR #140, opened 2024-03-11
Read average azimuth pixel spacing from S1 annotation files

gshiroma:
This PR updates the S1-reader to parse the average azimuth pixel spacing from S1 annotation files and storing it as the `Sentinel1BurstSlc` class attribute `average_azimuth_pixel_spacing`. This field is required by CEOS ARD specifications for Normalized Radar Backscatter (NRB) products (e.g., the OPERA RTC-S1 product).

The PR also remove repeated code that reads the field `imageAnnotation/imageInformation/numberOfSamples` from S1 annotation files.

### DIFF
--- a/src/s1reader/version.py
+++ b/src/s1reader/version.py
@@ -4,6 +4,7 @@ import collections
 # release history
 Tag = collections.namedtuple('Tag', 'version date')
 release_history = (
+    Tag('0.2.4', '2024-03-11'),
     Tag('0.2.3', '2023-09-21'),
     Tag('0.2.2', '2023-09-08'),
     Tag('0.2.1', '2023-08-23'),
--- a/src/s1reader/s1_annotation.py
+++ b/src/s1reader/s1_annotation.py
@@ -454,9 +454,6 @@ class ProductAnnotation(AnnotationBase):
         cls.number_of_samples = \
             cls._parse_scalar('imageAnnotation/imageInformation/numberOfSamples',
                               'scalar_int')
-        cls.number_of_samples = \
-            cls._parse_scalar('imageAnnotation/imageInformation/numberOfSamples',
-                              'scalar_int')
         cls.range_sampling_rate = \
             cls._parse_scalar('generalAnnotation/productInformation/rangeSamplingRate',
                               'scalar_float')
--- a/src/s1reader/s1_burst_slc.py
+++ b/src/s1reader/s1_burst_slc.py
@@ -218,6 +218,7 @@ class Sentinel1BurstSlc:
     radar_center_frequency: float
     wavelength: float
     azimuth_steer_rate: float
+    average_azimuth_pixel_spacing: float
     azimuth_time_interval: float
     slant_range_time: float
     starting_range: float
--- a/src/s1reader/s1_reader.py
+++ b/src/s1reader/s1_reader.py
@@ -835,6 +835,7 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
         orbit_direction = product_info_element.find('pass').text
 
         image_info_element = tree.find('imageAnnotation/imageInformation')
+        average_azimuth_pixel_spacing = float(image_info_element.find('azimuthPixelSpacing').text)
         azimuth_time_interval = float(image_info_element.find('azimuthTimeInterval').text)
         slant_range_time = float(image_info_element.find('slantRangeTime').text)
         ascending_node_time_annotation =\
@@ -1014,7 +1015,8 @@ def burst_from_xml(annotation_path: str, orbit_path: str, tiff_path: str,
         burst_misc_metadata = swath_misc_metadata.extract_by_aztime(sensing_start)
 
         bursts[i] = Sentinel1BurstSlc(ipf_version, sensing_start, radar_freq, wavelength,
-                                      azimuth_steer_rate, azimuth_time_interval,
+                                      azimuth_steer_rate, average_azimuth_pixel_spacing,
+                                      azimuth_time_interval,
                                       slant_range_time, starting_range, iw2_mid_range,
                                       range_sampling_rate, range_pxl_spacing,
                                       (n_lines, n_samples), az_fm_rate, doppler,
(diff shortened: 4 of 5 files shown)


seongsujeong (commented):
It looks like the major point of the change is to provide the azimuth spacing in meters instead of seconds. I agree that the spacing in meters would be more helpful for the users, but I would like to hear from @hfattahi about this change.

The most recent [CEOS ARD SAR spec](https://ceos.org/ard/files/PFS/SAR/v1.0/CEOS-ARD_PFS_Synthetic_Aperture_Radar_v1.0.pdf) does mention to include azimuth spacing in `1.6.7`, but does not specify what unit to use. Based on several factors, I would like to minimize the changes on the source code unless it is absolutely necessary.

gshiroma:
> It looks like the major point of the change is to provide the azimuth spacing in meters instead of seconds. I agree that the spacing in meters would be more helpful for the users, but I would like to hear from @hfattahi about this change.
> 
> The most recent [CEOS ARD SAR spec](https://ceos.org/ard/files/PFS/SAR/v1.0/CEOS-ARD_PFS_Synthetic_Aperture_Radar_v1.0.pdf) does mention to include azimuth spacing in `1.6.7`, but does not specify what unit to use. Based on several factors, I would like to minimize the changes on the source code unless it is absolutely necessary.

Thank you, @seongsujeong ! I agree. Let's wait for @hfattahi 's opinion on this.

src/s1reader/version.py
```diff
@@ -4,6 +4,7 @@
 # release history
 Tag = collections.namedtuple('Tag', 'version date')
 release_history = (
+    Tag('0.2.4', '2024-03-11'),
```
seongsujeong, line 7: The current RTC-S1 SAS uses s1reader `0.2.2`. Maybe we should think about a version number like `0.2.2.1` to minimize the changes in the patch release.

seongsujeong (approved):
LGTM. Thanks for the patch.

Aside from the code, my suggestion about the release is that we might need to do something similar to what we're doing on ISCE3 for CSLC-S1 patch release: it would be better to branch out from `0.2.2`, apply the patch, and make use of that version for the patch release for RTC-S1. The version for `s1reader` would be something like `0.2.2.1`. That is not a typical nomenclature for the versioning for OPERA, but it's not my first time to see such format in other projects.

I think cherrypicking will to the trick, but I'll try out few options after getting this PR merged to the main branch.

src/s1reader/version.py
hfattahi: Let's keep the current versioning scheme please.

hfattahi:
> LGTM. Thanks for the patch.
> 
> Aside from the code, my suggestion about the release is that we might need to do something similar to what we're doing on ISCE3 for CSLC-S1 patch release: it would be better to branch out from `0.2.2`, apply the patch, and make use of that version for the patch release for RTC-S1. The version for `s1reader` would be something like `0.2.2.1`. That is not a typical nomenclature for the versioning for OPERA, but it's not my first time to see such format in other projects.
> 
> I think cherrypicking will to the trick, but I'll try out few options after getting this PR merged to the main branch.

@seongsujeong I understand your concern but in this case there has not been any changes to s1reader since the previous opera delivery. So I think we are not imposing unwanted changes here for a patch release. For example see [here](https://github.com/opera-adt/COMPASS/blob/main/docker/Dockerfile#L47) where we use s1reader 0.2.3. So I think we should be fine to bump the version up to 0.2.4. What do you think?

hfattahi (approved):
LGTM. Thanks @gshiroma

gshiroma:
> LGTM. Thanks @gshiroma

Thank you, @hfattahi , for reviewing this PR! I talked to @seongsujeong offline and his biggest concern is the "polygon issue" reported by @scottstanie [here](https://github.com/isce-framework/s1-reader/issues/118). All the datasets reported by @scottstanie are from 2015, so that shouldn't affect forward production. Also, from a first verification, it seems that the handling those bad datasets will be the same, i.e, raising and error and stop the processing. But I'd like to run more tests today to make sure everything is OK.

seongsujeong:
@hfattahi @gshiroma Thanks for the feedback. Here is a bit more context about the versioning
- After the discussion, we decided to let the SAS fail when we cannot parse the polygon from the S1 IW annotation files
- But the way to make the software fail is different between RTC-S1 and CSLC-S1
   - RTC-S1 is using s1reader `0.2.2`. The SAS failure takes place in `s1reader` side
   - CSLC-S1 is using s1reader `0.2.3`. The SAS failure takes place in `COMPASS` side. This version fixes the missing data issue for azimuth FM rate data, and the polygon (it will return empty geometry)
   - The reasoning behind returning the empty polygon was because we wanted to read the S1 IW SLC bursts regardless of the missing metadata, and do something other than the OPERA L2 workflow.


Below are the solutions I can think about:
1. Merge this PR, bump the version to `0.2.4` and let RTC-S1 use that
1.1. We will have to modify RTC-S1 to let the SAS fail

2. Post another PR to let `s1reader` 
2.1. That will make the reader fail to load some old SAFE files [example here](https://github.com/isce-framework/s1-reader/issues/118). 

3. Use derivative version number like I've suggested above
3.1. Probably the simplest solution

Please let us know what you think

hfattahi:
> @hfattahi @gshiroma Thanks for the feedback. Here is a bit more context about the versioning
> 
> * After the discussion, we decided to let the SAS fail when we cannot parse the polygon from the S1 IW annotation files
> * But the way to make the software fail is different between RTC-S1 and CSLC-S1
>   
>   * RTC-S1 is using s1reader `0.2.2`. The SAS failure takes place in `s1reader` side
>   * CSLC-S1 is using s1reader `0.2.3`. The SAS failure takes place in `COMPASS` side. This version fixes the missing data issue for azimuth FM rate data, and the polygon (it will return empty geometry)
>   * The reasoning behind returning the empty polygon was because we wanted to read the S1 IW SLC bursts regardless of the missing metadata, and do something other than the OPERA L2 workflow.
> 
> Below are the solutions I can think about:
> 
> 1. Merge this PR, bump the version to `0.2.4` and let RTC-S1 use that
>    1.1. We might need to modify RTC-S1 to let the SAS fail
> 2. Post another PR to let `s1reader`
>    2.1. That will make the reader fail to load some old SAFE files [example here](https://github.com/isce-framework/s1-reader/issues/118).
> 3. Use derivative version number like I've suggested above
>    3.1. Probably the simplest solution
>    3.2. Might need to branch out from the tag `v0.2.2`, cherrypick the commits in this PR to that branch
> 
> Please let us know what you think

Thanks @seongsujeong for pointing out those details. I'm ok with your Option-1.

PR #142, opened 2025-02-06
Fix `s1_orbit` to use ASF S3 orbit source

### LINKED ISSUE #3: Abiltiy to download orbits from ASF

scottstanie:
Scihub has been broken, leading to broken functionality for `s1_orbit1`. This is also breaking local COMPASS use due to https://github.com/opera-adt/COMPASS/issues/226

You should be able to run (random SAFE name):
```python
import s1reader.s1_orbit

orbit_path =  s1reader.s1_orbit.get_orbit_file_from_dir("S1A_IW_SLC__1SDV_20240902T233705_20240902T233732_055493_06C53E_4976", ".", auto_download=True)
```
and get `S1A_OPER_AUX_POEORB_OPOD_20240922T070612_V20240901T225942_20240903T005942.EOF`

This takes the idea from https://github.com/scottstanie/sentineleof/pull/66 to use the new ASF public orbit source.

Removed:
• The global Scihub credentials
• Functions that queried the Copernicus GNSS API (check_internet_connection, _get_orbit_dict, _download_orbit_file)

Added/Updated:
- We no longer need to have an existing .zip file to run this (I'm not sure why we had the `exist` check: we just parse the name for start/stop dates, so we don't need an actual file.
- A new S3-based listing function (list_public_bucket()) and orbit file getter (get_orbit_files())
- A helper to download orbit files from ASF S3 (download_orbit_file_from_s3())
- An updated retrieve_orbit_file() that works with the S3 service, including fallback to handling pair concatenation if needed
This should “fix” the broken module by removing the dead Scihub parts while still providing the previous functionality using the new ASF S3 service.
- `tests/requirements.txt` now list this the pip requirements for running tests (`pytest`, and new `pytest-recording`)
- Added the `pytest-recording` so that we save the API result for the function which grabs the list of orbits from S3. This was generated with `pytest --record-mode all tests/` 

I've tried to keep as much old stuff as possible, but I'm not 100% clear which portions are necessary for current OPERA production and which are handled by the PCM staging logic. I'm most unsure around the concatenation logic.

Closes #3


tests/test_orbit.py
```diff
@@ -137,7 +159,7 @@ def test_anx_time(test_paths):
 def test_combine_xml_orbit_elements(tmp_path, test_paths):
     slc_file = (
         tmp_path
-        / "S1A_IW_SLC__1SDV_20230823T154908_20230823T154935_050004_060418_521B.SAFE"
+        / "S1A_IW_SLC__1SDV_20230823T154908_20230823T154935_050004_060418_521B.SAFES1A_IW_SLC__1SDV_20230823T154908_20230823T154935_050004_060418_521B"
```
scottstanie, line 162:
```suggestion
        / "S1A_IW_SLC__1SDV_20230823T154908_20230823T154935_050004_060418_521B.SAFE"
```

src/s1reader/s1_orbit.py
```diff
@@ -3,513 +3,442 @@
...
+    else:
+        return [local_file_earlier, local_file_later]
 
 
 def _parse_safe_filename(safe_filename):
-    """
-    Extract info from S1-A/B SAFE filename
-    SAFE filename structure:
-    S1A_IW_SLC__1SDV_20150224T114043_20150224T114111_004764_005E86_AD02.SAFE
+    """Extract info from S1-A/B SAFE filename.
```
seongsujeong, line 282:
```suggestion
    """Extract info from S1-A/B/C SAFE filename.
```

src/s1reader/s1_orbit.py
```diff
@@ -3,513 +3,442 @@
...
 
-    Parameters
-    -----------
-    safe_filename: string
-       Path to S1-A/B SAFE file
+    SAFE filename structure:
```
seongsujeong, line 284:
```suggestion
    SAFE filename example:
```

src/s1reader/s1_orbit.py
```diff
@@ -560,66 +490,71 @@ def get_resorb_pair_from_list(
...
     # as a concatenated RESORB file
+    # Either concatenate the orbit, or return the RESORB pair
+    # concatenate the RESORB xml file.
+    # NOTE Careful about the order how the RESORBs are concatenated to avoid
+    # the non-uniform spacing of OSVs during the sensing times
+    # 11111111111111111111111                                    2222222222222222222222
+    #                2222222222222222222222      11111111111111111111111
+    # 1111111111111112222222222222222222222      11111111111111111111111222222222222222
+    #  |              |---sensing time---|        |               |---sensing time---|
+    # ANX crossing                               ANX crossing
```
seongsujeong, line 510:
```suggestion
    # secondary RESORB    11111111111111111111111                                    2222222222222222222222
    # reference RESORB                   2222222222222222222222      11111111111111111111111
    # concatenated        1111111111111112222222222222222222222      11111111111111111111111222222222222222
    #                      |              |---sensing time---|        |               |---sensing time---|
    #                      ANX crossing                               ANX crossing
```

tests/test_orbit.py
```diff
@@ -4,24 +4,46 @@
...
+import s1reader.s1_orbit
 
 
-def test_get_orbit_file(test_paths):
-    """
-    Unit test for `get_orbit_file_from_dir`
-    """
+@pytest.mark.vcr
+def test_list_public_bucket_resorb():
+    resorbs = list_public_bucket(ASF_BUCKET_NAME, prefix="AUX_RESORB")
```
seongsujeong, line 30: Looks like the test code here attempts to download RESORB from `s3://s1-orbits` bucket. Please note that the bucket only keeps the last 90 days of orbit files as described in the documentation [here](https://registry.opendata.aws/s1-orbits/)

seongsujeong (commented):
Here are my comments. Most of them are minor.

tests/test_orbit.py
scottstanie: I can add a note to `tests/README` about how this works- 

In general, we want the unit tests to run without requiring external API access. They way this occurs:

- The `@pytest.mark.vcr` decorator uses [the pytest-recording library](https://github.com/kiwicom/pytest-recording)
- When you run `pytest --record-mode all` the first time locally, it will run the actual code including the network request. 
- The results all get saved in those YAML files in the `tests/cassettes` folder: they are long because they saved all the HTTP traffic that occurred, including the resulting data returned.
- Each subsequent run of `pytest` mocks the `requests.get` calls, and instead reads from the YAML files.

See also https://github.com/kevin1024/vcrpy?tab=readme-ov-file#rationale for a succinct rationale.

seongsujeong (approved):
LGTM. Thanks for the PR!

